PR #149, opened 2018-09-21
Convert TimerJsInterface to use seconds

jsundra:
Noticed `setTimeout` was still using milliseconds, while `time.*` uses seconds. A bit of a deviation from `setTimeout`'s traditional web usage, but better to be standardized across our APIs!


jsundra:
Note: Updating this might be tricky since a certain untethered experience relies on `setTimeout` with milliseconds right now

thegoldenmule:
I understand the idea (since the `time` api is seconds), but `setTimeout` is a standard JS API that uses milliseconds.

PR #150, opened 2018-09-24
Feature/asset replacement

thegoldenmule:
Work to support asset replacement.

### DIFF
--- a/Assets/Source/Application/Services/AssetUpdateService.cs
+++ b/Assets/Source/Application/Services/AssetUpdateService.cs
@@ -180,10 +180,21 @@ namespace CreateAR.EnkluPlayer
                 return;
             }
 
-            asset.Uri = string.Format(
-                "{0}_{1}.bundle",
-                asset.Uri.Substring(0, index),
-                GetBuildTarget());
+            if (asset.Uri.Contains("{{platform}}"))
+            {
+                // v2
+                asset.Uri = asset.Uri.Replace(
+                    "{{platform}}",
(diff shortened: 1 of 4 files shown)


Assets/Source/Application/Services/AssetUpdateService.cs
```diff
@@ -180,10 +180,21 @@ private void FormatAssetData(AssetData asset)
...
+                    GetBuildTarget());
+            }
+            else
+            {
+                // v1
+                asset.Uri = string.Format(
+                    "{0}_{1}.bundle",
+                    asset.Uri.Substring(0, index),
+                    GetBuildTarget());
+            }
```
thegoldenmule: An unfortunate necessity: the asset server now generates uris differently. I would like to make a db change eventually so we can rip this out.

Assets/Source/Player/Asset/StandardAssetLoader.cs
```diff
@@ -148,5 +147,21 @@ private IEnumerator ProcessQueue()
...
+        /// <returns></returns>
+        private static string AssetName(AssetData data)
+        {
+            var path = Path.GetFileNameWithoutExtension(data.Uri);
+            if (string.IsNullOrEmpty(path))
+            {
+                return string.Empty;
+            }
+
+            return path.Split('_')[0];
```
thegoldenmule: Explanation: the client used to assume the main asset inside of the bundle was named after the id, but now there is a separate guid that the asset server names the asset.

PR #153, opened 2018-09-25
EK-413 Proximity RTE spam

jsundra:
When exiting an experience or entering edit mode, if the proximity js api had been used the runtime would be spammed with null refs and become unbearably slow.

Two issues:
1. ProximityManager unsubscribed from schema changes on Element destruction, but never updated ProximityChecker to remove tracking.
- ProximityManager now has a `RemoveElement` method that gets called on Element destruction.

2. Since ProximityChecker now uses GameObjects, OnDestroyed would be invoked after Widget instances already removed their GameObject reference, and not be able to remove tracking properly.
- Element invokes `OnDestroyed` before calling its internal unload methods. This seems like a good change so that anything relying on an Element's state can cleanup properly before the Element starts cleaning itself up. The current usage of `OnDestroyed`s seems safe to change its order as well.

### DIFF
--- a/Assets/Source/Player/IUX/Element/Element.cs
+++ b/Assets/Source/Player/IUX/Element/Element.cs
@@ -212,6 +212,11 @@ namespace CreateAR.EnkluPlayer.IUX
         /// </summary>
         public void Destroy()
         {
+            if (OnDestroyed != null) {
+                OnDestroyed(this);
+                OnDestroyed = null;
+            }
+
             UnloadInternalBeforeChildren();
 
             // destroy children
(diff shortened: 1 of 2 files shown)


Assets/Source/Player/IUX/Element/Element.cs
```diff
@@ -212,6 +212,12 @@ public void LateFrameUpdate()
         /// </summary>
         public void Destroy()
         {
+            if (OnDestroyed != null)
+            {
+                OnDestroyed(this);
+                OnDestroyed = null;
+            }
+
```
thegoldenmule: Hmmm-- do tests pass? This may have unintended side-effects.

jsundra: They do! Well, an unrelated jsapi one that broke earlier is failing. But all the element ones still work.

Definitely agree that it look like a scary change though, ripe with side effects. The seemingly biggest impact is that `ElementControllerGroup` calls its `Element_OnDestroyed` earlier in the destruction/unload cycle from before- but all that does is queue up a GameObject for deletion.

PR #155, opened 2018-09-26
Schema changes don't invoke watchers if the value was the same

jsundra:
When playing with some things today I realized there were some duplicate invokes going out. Probably more common with the web builds than a holo build since there's the update messages & preview messages, but this doesn't seem like a bad check to have in place?

I used this change with all of my normal work today and didn't run into any bugs, but haven't done any exhaustive testing by any means.

### DIFF
--- a/Assets/Source/Player/IUX/Element/ElementSchemaProp.cs
+++ b/Assets/Source/Player/IUX/Element/ElementSchemaProp.cs
@@ -73,6 +73,17 @@ namespace CreateAR.EnkluPlayer.IUX
             }
             set
             {
+                // Early out if the same value is being set
+                if (_value == null && value == null)
+                {
+                    return;
+                }
+
+                if (_value != null && _value.Equals(value))
+                {
+                    return;
+                }
+
                 // break connection
                 if (null != _parent)
                 {


Assets/Source/Player/IUX/Element/ElementSchemaProp.cs
```diff
@@ -73,6 +73,17 @@ public T Value
...
+                // Early out if the same value is being set
+                if (_value == null && value == null)
+                {
+                    return;
+                }
+
+                if (_value != null && _value.Equals(value))
+                {
+                    return;
+                }
```
thegoldenmule: Do tests pass? I can think of one edge case: a connection with a parent will not be broken even though you are setting a new value. This is probably desired, however.

jsundra: Yep, tests passed. Hmm... if the parent connection breaking is desired, that logic could still happen on redundant sets, and just the change invoke be throttled still. Haven't gone too far into the schema linkage though to know side effects off the top of my head.

thegoldenmule: I shouldn't have merged-- sorry about that! :D

PR #156, opened 2018-09-26
ScreenWidget smoothing

jsundra:
When using a ScreenWidget for a "try and use a gesture!" screen (or anything that might hang around for a bit), the jitter from slight head movements is pretty noticeable. To try and make this UX a bit better, I added some schema to control if/how a ScreenWidget should reposition itself.

It isn't perfect, and despite having some frame rate independence, it can be jerky at times if there's a lot going on in the scene. But, playing with the values a bit, the results seem good in my extremely biased opinion.

Before:
screenbefore

After:
screenafter

### DIFF
--- a/Assets/Source/Math/Vec3.cs
+++ b/Assets/Source/Math/Vec3.cs
@@ -8,6 +8,9 @@ namespace CreateAR.EnkluPlayer
     [Serializable]
     public struct Vec3
     {
+        /// <summary>
+        /// Radians -> Degrees.
+        /// </summary>
         private const float RAD2_DEG = (float) (360 / (2 * Math.PI));
 
         /// <summary>
--- a/Assets/Source/Math/Vec3.cs
+++ b/Assets/Source/Math/Vec3.cs
@@ -8,6 +8,8 @@ namespace CreateAR.EnkluPlayer
     [Serializable]
     public struct Vec3
     {
+        private const float RAD2_DEG = (float) (360 / (2 * Math.PI));
+
         /// <summary>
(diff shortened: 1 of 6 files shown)


Assets/Source/Math/Vec3.cs
```diff
@@ -8,6 +8,8 @@ namespace CreateAR.EnkluPlayer
     [Serializable]
     public struct Vec3
     {
+        private const float RAD2_DEG = (float) (360 / (2 * Math.PI));
```
thegoldenmule: Missing comment.

Assets/Source/Player/IUX/Widget/ScreenWidget.cs
```diff
@@ -37,16 +74,36 @@ protected override void LoadInternalAfterChildren()
             base.LoadInternalAfterChildren();
 
             _distanceProp = Schema.GetOwn("distance", 1f);
+            _stabilizationProp = Schema.GetOwn("stabilization", 2f);
+            _smoothingProp = Schema.GetOwn("smoothing", 15f);
```
thegoldenmule: I would rather these defaults go into `ElementFactory` than here. In addition, please add new props to the doc: https://github.com/enklu/enkluplayer/blob/master/docs/element.schema.properties.md.

jsundra: 👍 I'll shift the defaults up and add to docs. I guess the usage should probably switch to `Get` instead of `GetOwn` too.

PR #165, opened 2018-09-28
EK-451 Activator Buffers

jsundra:
- Adds a new Activator state: `ActivatorImminentState`. Not all that in love with the naming.
This state essentially takes the responsibilities `ActivatorReadyState` had. it transitions to Activating when needed, and de-progresses a canceled "activating" as needed. The ready state now is just idle. In the future we can style ready & close differently if we'd like.

- Adds a new Raycast overload to `IRaycaster`.

- Changes Aim range from 0 -> 1 to -1 -> 1.
The new area in the range signifies how close the user is to the activator while in the buffer area.

- `IntentionManager` no longer early outs.
To handle large buffers & overlapping buttons, IntentionManager no longer early outs when it finds a valid raycast. Instead it continues raycasting, and then picks the Interactable with the closest angular offset.

### DIFF
--- a/Assets/Source/Player/IUX/Activator/ActivatorPrimitive.cs
+++ b/Assets/Source/Player/IUX/Activator/ActivatorPrimitive.cs
@@ -457,20 +457,37 @@ namespace CreateAR.EnkluPlayer.IUX
             var directionToButton = delta.Normalized;
 
             var eyeDistance = delta.Magnitude;
-            var radius = _renderer.Radius;
+            var activateRadius = _renderer.ActivateRadius;
+            var focusRadius = _renderer.FocusRadius;
 
-            var maxTheta = Mathf.Atan2(radius, eyeDistance);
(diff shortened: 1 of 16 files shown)


Assets/Source/Player/IUX/Activator/ActivatorPrimitive.cs
```diff
@@ -254,17 +258,27 @@ public float IconScale
 
         /// <inheritdoc cref="IRaycaster"/>
         public bool Raycast(Vec3 origin, Vec3 direction)
+        {
+            Vec3 hitPosition, colliderCenter;
+            return Raycast(origin, direction, out hitPosition, out colliderCenter);
+        }
+
+        public bool Raycast(Vec3 origin, Vec3 direction, out Vec3 hitPosition, out Vec3 colliderCenter)
```
thegoldenmule: Comments.

PR #170, opened 2018-10-02
Feature/snaps trigger

thegoldenmule:
Opening this for early review, but not quite ready to go yet.

* Stubbed out a snap API for JS.
* Refactored device registration a bit.
* Added button to re-sync device registration.
* Added current app version to edit menu.
* Added app name to play and design menus.


Assets/Source/Messages/MessageTypes.cs
```diff
@@ -35,6 +35,7 @@ public static class MessageTypes
         public const int DEVICE_REGISTRATION_COMPLETE = 1000004;
         public const int ANCHOR_AUTOEXPORT = 1000005;
         public const int ANCHOR_RESETPRIMARY = 1000006;
+        public const int DEVICE_REGISTRATION = 1000005;
```
jsundra: This ID conflicts with `ANCHOR_AUTOEXPORT`

Assets/Source/Player/Scripting/Interop/SnapJsInterface.cs
```diff
@@ -0,0 +1,75 @@
...
+            _preferences = prefs;
+        }
+
+        /// <summary>
+        /// Triggers a snap to be taken.
+        /// </summary>
+        public void trigger()
+        {
+            _preferences
+                .ForCurrentUser()
```
jsundra: Maybe this should be done when the interface inits and cached? Though it probably wouldn't take that long to resolve so it probably doesn't matter.

I guess to _really_ do that right as well, there'd probably need to be a way to refresh jsinterfaces, so yeah maybe just food for thought.

thegoldenmule: It shouldn't cache, it's not a user prefs object, it's a special wrapper that helps synchronize it.

PR #174, opened 2018-10-04
Enhancement/anchor bypass

thegoldenmule:
* Anchor bypass.
* Version check bypass fix.
* Relative positioning for failed anchors.


Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
```diff
@@ -509,94 +534,200 @@ private void OpenStatusUI()
...
+                        var B = Matrix4x4.TRS(locatedPos, locatedRot, Vector3.one);
+
+                        // T * A = B
+                        T = B * Matrix4x4.Inverse(A);
+                    }
+
+                    // place non-located anchors relative to the located anchor
+                    for (int i = 0, len = _anchors.Count; i < len; i++)
+                    {
+                        var anchor = _anchors[i];
```
jsundra: Minor, but there could probably be a `if (anchor == located) continue` skip, since that case would be positioning it relative to itself.

PR #176, opened 2018-10-10
MaterialJsApi & AudioJsApi

jsundra:
Finally got around to testing this- new MaterialJsApi lets scripts change shader parameters on the fly.

For now, everything uses the renderer's shared material. Eventually we can add support for optionally splitting into unique instances, etc. Also, I didn't want to rush and make a `Vec4` class and/or expose `Col4` to scripting yet, so I added a quick overload to allow changing a color + alpha if needed.

Also, added an AudioJsApi for modifying an AudioSource attached to an uploaded asset.

### DIFF
--- a/Assets/Source/Player/Scripting/Interface/Elements/AudioJsApi.cs
+++ b/Assets/Source/Player/Scripting/Interface/Elements/AudioJsApi.cs
@@ -145,5 +145,23 @@ namespace CreateAR.EnkluPlayer
         {
             _audio.maxDistance = distance;
         }
+
+        /// <summary>
+        /// Gets the doppler level.
+        /// </summary>
+        /// <returns></returns>
+        public float getDoplerLevel()
+        {
+            return _audio.dopplerLevel;
+        }
+
+        /// <summary>
+        /// Sets the doppler level.
+        /// </summary>
+        /// <param name="doppler"></param>
+        public void setDoplerLevel(float doppler)
+        {
+            _audio.dopplerLevel = doppler;
+        }
     }
 }
--- a/Assets/Source/Player/Scripting/Interface/Elements/AudioJsApi.cs
+++ b/Assets/Source/Player/Scripting/Interface/Elements/AudioJsApi.cs
@@ -1,0 +1,149 @@
+﻿using UnityEngine;
+
+namespace CreateAR.EnkluPlayer
+{
+    /// <summary>
+    /// An interface to Unity's AudioSource.
+    /// </summary>
+    public class AudioJsApi
+    {
+        /// <summary>
+        /// Underlying AudioSource to wrap.
+        /// </summary>
+        private readonly AudioSource _audio;
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="audio"></param>
+        public AudioJsApi(AudioSource audio)
+        {
+            _audio = audio;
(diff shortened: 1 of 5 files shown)


Assets/Source/Player/Scripting/Interface/Elements/AudioJsApi.cs
```diff
@@ -0,0 +1,167 @@
...
+        public float getDoplerLevel()
+        {
+            return _audio.dopplerLevel;
+        }
+
+        /// <summary>
+        /// Sets the doppler level.
+        /// </summary>
+        /// <param name="doppler"></param>
+        public void setDoplerLevel(float doppler)
```
thegoldenmule: * Doppler

Assets/Source/Player/Scripting/Interface/Elements/AudioJsApi.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        /// <summary>
+        /// Sets the volume.
+        /// </summary>
+        /// <param name="volume"></param>
+        public void setVolume(float volume)
+        {
+            _audio.volume = volume;
+        }
```
thegoldenmule: Why not make all of these properties? `audio.volume = 3;` seems like a better interface.

Assets/Source/Player/Scripting/Interface/Elements/ContentElementJs.cs
```diff
@@ -15,6 +15,13 @@ public class ContentElementJs : ElementJs
         /// </summary>
         public AnimatorJsApi animator { get; private set; }
 
+        /// <summary>
+        /// The material interface.
+        /// </summary>
+        public MaterialJsApi material { get; private set; }
+
+        public AudioJsApi audio { get; private set; }
```
thegoldenmule: Comments.

Assets/Source/Player/Scripting/Interface/Elements/AudioJsApi.cs
jsundra: Hmm, I originally did it just to be more like the Material & Animator interfaces, which only use functions because their wrapping Unity interface act that way. But there's no reason it has to be similar to those, so yeah I'll switch them to be accessors because that does feel better.

thegoldenmule: :+1:

Assets/Source/Player/Scripting/Interface/Elements/MaterialJsApi.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        /// <summary>
+        /// Sets a float material property.
+        /// </summary>
+        /// <param name="param"></param>
+        /// <param name="value"></param>
+        public void setFloat(string param, float value)
+        {
+            _renderer.sharedMaterial.SetFloat(param, value);
```
thegoldenmule: Can `sharedMaterial` be null?

jsundra: Good catch! Ran a quick test and it can be. Added a safety to `ContentElementJs` to ensure not just a `Renderer` is found, but that `sharedMaterial` isn't null too.

PR #178, opened 2018-10-11
Bugfix/softlaunch 1

thegoldenmule:
Many metrics and some optimizations.


Assets/Source/Application/Util/DeviceHelper.cs
```diff
@@ -15,6 +15,19 @@ public static bool IsHoloLens()
...
 #else
             return false;
+#endif
+        }
+
+        /// <summary>
+        /// True iff the platform is WebGL.
+        /// </summary>
+        /// <returns></returns>
+        public static bool IsWebGl()
```
jsundra: OCD thing, but could this be `IsWebGL`?

Assets/Source/Player/Scripting/Interop/MetricsJsInterface.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                _timers[index].Stop(id);
+
+                _timerIds.RemoveAt(index);
+                _timers.RemoveAt(index);
+            }
+        }
+
+        /// <summary>
+        /// Aborsts a timer for a specific invocation.
```
jsundra: Aborsts!

Assets/Source/Player/Vine/ActionWorker.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// True iff alive.
+        /// </summary>
+        private bool _isAlive;
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        public ActionWorker(IBootstrapper bootstrapper)
+        {
+            bootstrapper.BootstrapCoroutine(Poll());
```
jsundra: I remember you saying you were going to try and throw vine parsing onto its own thread. Did you mean that they'd just be async instead? The locks in here all look good, but I don't think they're needed since coroutines still run on the main thread?

Assets/Source/Application/Util/DeviceHelper.cs
thegoldenmule: I would rather not. Our standard has been `Gl`, `Ios`, etc.

Assets/Source/Player/Scripting/Interop/MetricsJsInterface.cs
thegoldenmule: :+1:

Assets/Source/Player/Vine/ActionWorker.cs
thegoldenmule: That's not quite how this class works. `Start()` is called on a different thread. That other thread needs to synchronize with the main thread, so you'll see below how we stick results into the mailbox. The mailbox is polled on the main thread so that the callback is always called on the main thread.

PR #179, opened 2018-10-13
Col4 JS API.

thegoldenmule:


Assets/Source/Player/Scripting/Interface/Math/QuatMethods.cs
```diff
@@ -1,52 +1,52 @@
...
-
-        /// <summary>
-        /// Creates a Quat from Euler rotations.
-        /// </summary>
-        public Quat eul(float x, float y, float z)
-        {
-            return Quat.Euler(x, y, z);
-        }
-
+﻿namespace CreateAR.EnkluPlayer
```
thegoldenmule: This file had mixed line endings.

thegoldenmule:
Crap, you're right. I'll add the functions real quick.

thegoldenmule:
@jsundra Added!

PR #180, opened 2018-10-15
Feature/loadtimes

thegoldenmule:
* Added DelayedScriptCache.
* Play mode doesn't connect to Trellis.
* Added config options for periodic updates, i.e. no more than once every X minutes.
* Added config option to skip device registration.
* Added config option to skip version check.
* Some small perf fixes.

### DIFF
--- a/Assets/Source/Player/Scripting/Interop/ProximityChecker.cs
+++ b/Assets/Source/Player/Scripting/Interop/ProximityChecker.cs
@@ -114,7 +114,7 @@ namespace CreateAR.EnkluPlayer.Scripting
             var backingGameObject = EntityToGameObject(entity);
             if (backingGameObject == null)
             {
-                Log.Warning(this, "Can't find backing gameobject for " + entity);
+                Log.Info(this, "Can't find backing gameobject for " + entity);
                 return;
             }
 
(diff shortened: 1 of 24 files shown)


Assets/Source/Player/Scripting/Interop/ProximityChecker.cs
```diff
@@ -114,7 +114,7 @@ public void SetElementState(IEntityJs entity, bool isListening, bool isTrigger)
             var backingGameObject = EntityToGameObject(entity);
             if (backingGameObject == null)
             {
-                Log.Warning(this, "Can't find backing gameobject for " + entity);
+                Log.Info(this, "Can't find backing gameobject for " + entity);
```
jsundra: This log could actually go away - with things like transitions being in the scene, it's expected for this to happen in the current setup of all Elements potentially entering ProximityChecker

Assets/Source/Player/Scripting/Interop/ProximityChecker.cs
```diff
@@ -293,7 +293,7 @@ private GameObject EntityToGameObject(IEntityJs entity)
                 var widget = elementJs.Element as Widget;
                 if (widget == null) 
                 {
-                    Log.Warning(this, "ElementJs was not a Widget");
+                    Log.Info(this, "ElementJs was not a Widget");
```
jsundra: Ditto for this one, entirely valid code path currently.

jsundra:
Yeah, even with 10s scripts aren't loading properly (the biggest thing to notice is that the timers jsinterface doesn't seem to work). Nothing obvious in the console of course, something we'll have to look into further :/

thegoldenmule:
Can you try with passthrough then?

PR #182, opened 2018-10-15
Ensure vines finish running before behaviors start

jsundra:
Adding back this old guarantee. With this, scripting no longer needs weak safeties such as
```
function enter(){
  timers.setTimeout(lateInit.bind(this), 3000);
}

function lateInit(){
  vineElement = this.findOne('..id');
}
```

Tested on a demo scene with a handful of vines/scripts loading side by side.


thegoldenmule:
Excellent. Looks good to me. Eventually we'll need to keep all the tokens from the vine creation and abort them if we get a script update while they are still being setup.

PR #186, opened 2018-10-16
Add `system.device` to scripting

jsundra:
The scripting api itself isn't bad- but getting `IDeviceMetaProvider` into `SystemJsApi` was kind of gross. I'm thinking in the future, we might want to have some interface lookup/resolver that is responsible for managing instances rather than `ScriptCollectionRunner` crudely creating a single instance on first run :/

Open to ideas of how to make this better, was trying to balance time/profit.

### DIFF
--- a/Assets/Source/Player/IUX/Content/ScriptCollectionRunner.cs
+++ b/Assets/Source/Player/IUX/Content/ScriptCollectionRunner.cs
@@ -85,6 +85,12 @@ namespace CreateAR.EnkluPlayer.Scripting
         private readonly List<VineMonoBehaviour> _vineComponents = new List<VineMonoBehaviour>();
 
         /// <summary>
+        /// Cached/shared SystemJsApi instance.
+        /// TODO: Make an IScriptInterfaceResolver that's part of IScriptManager maybe?!
+        /// </summary>
+        private static SystemJsApi _systemJsApi;
+
+        /// <summary>
         /// Constructor.
         /// </summary>
         public ScriptCollectionRunner(
@@ -92,6 +98,7 @@ namespace CreateAR.EnkluPlayer.Scripting
             IScriptRequireResolver resolver,
             IElementJsCache jsCache,
             IElementJsFactory elementJsFactory,
+            IDeviceMetaProvider deviceMetaProvider,
             GameObject root,
             Element element)
         {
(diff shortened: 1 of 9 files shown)


Assets/Source/Player/IUX/Content/ScriptCollectionRunner.cs
```diff
@@ -101,6 +108,11 @@ private enum SetupState
             _elementJsFactory = elementJsFactory;
             _root = root;
             _element = element;
+
+            if (_systemJsApi == null)
+            {
+                _systemJsApi = new SystemJsApi(deviceMetaProvider);
+            }
```
thegoldenmule: Haaaah! This seems particularly evil. I would suggest just using a static reference to `SystemJsApi`.

jsundra:
Pretty sure there's only just this 1 instance since I removed its self-creating instance. But where would you suggest constructing the static reference from? The other thought I had was to leave it managing its own static instance still and referencing that- but then calling something like `SystemJsApi.SetDeviceMetaProvider(x)`, but that also seemed gross :/

thegoldenmule:
What about something like this inside `EnkluPlayerModule`:

```
SystemJsApi.Instance.DeviceMetaProvider = binder.GetInstance<IDeviceMetaProvider>();
```

It's a tad ugly, but setting up these big objects inside the IoC container is fairly contained. Then in `UnityScriptingHost`:

```
SetValue('system', SystemJsApi.Instance);
```

jsundra:
I like it! Seems like a pretty good way to keep the configuration high level.

Assets/Source/Player/Scripting/Interface/App/SystemJsApi.cs
```diff
@@ -8,7 +10,26 @@ public class SystemJsApi
...
         /// The instance.
         /// </summary>
-        public static SystemJsApi Instance = new SystemJsApi();
+        public static SystemJsApi Instance = new SystemJsApi();
+
+        /// <summary>
+        /// Provider for getting DeviceMeta.
+        /// </summary>
+        public static IDeviceMetaProvider DeviceMetaProvider {
+            set {
```
thegoldenmule: Braces on separate lines.

thegoldenmule:
COOL

PR #187, opened 2018-10-16
Performance/jint

thegoldenmule:
Some past work on JS garbage generation. In the V3 scene I'm seeing 50-70 kb less garbage _per frame_.


Assets/Source/Player/IUX/Content/ScriptCollectionRunner.cs
```diff
@@ -212,6 +217,13 @@ public void Teardown()
                 _caches[i].Clear();
             }
             _caches.Clear();
+
+            // destroy engines
+            for (int i = 0, len = _hosts.Count; i < len; i++)
+            {
+                _hosts[i].Destroy();
+            }
+            _hosts.Clear();
```
jsundra: Is the list of hosts populated?

Assets/Source/Util/OptimizedObjectPool.cs
```diff
@@ -0,0 +1,207 @@
...
+                    return default(T);
+                }
+            }
+
+            // pick available
+            var index = _availableIndices.Dequeue();
+            var instance = _allocated[index];
+            instance.Available = false;
+            return instance;
+        }
```
jsundra: `Get` seems a little counter to how I would have thought a pool would work. If a pool is intended to be dynamic, that makes sense- but for a static pool, wouldn't you want things to blow up/error so that memory pain points can be identified where more objects are being created/requested than expected?

Assets/ThirdParty/Jint/Jint/Engine.cs
```diff
@@ -220,15 +228,14 @@ public Engine(Action<Options> options)
...
-                {
-                    LexicalEnvironment = lexicalEnvironment,
-                    VariableEnvironment = variableEnvironment,
-                    ThisBinding = thisBinding
-                };
-            _executionContexts.Push(executionContext);
+            var context = _poolExecutionContext.Get();
+            context.LexicalEnvironment = lexicalEnvironment;
+            context.VariableEnvironment = variableEnvironment;
+            context.ThisBinding = thisBinding;
```
jsundra: Just curious for the change in notation - any special reason?

Assets/Source/Util/OptimizedObjectPool.cs
thegoldenmule: This pool can be dynamic or static, given by the input parameters. In general, I don't think it's necessarily useful for the pool itself to throw an exception-- the caller is free to throw one if if null is returned. Otherwise, you'd always need a try/catch around `Get`.

Assets/ThirdParty/Jint/Jint/Engine.cs
thegoldenmule: Not sure what you're referring to. I didn't make any of these object or name them, I'm just pooling them.

Assets/Source/Player/IUX/Content/ScriptCollectionRunner.cs
thegoldenmule: Whoops, doesn't look like it.

PR #189, opened 2018-10-16
Add extra, optional, "admin" guarding around specific voice commands.

jsundra:
Similar to the debug voice guard, there's a global config option to disable it. For individual voice registrations, an optional bool can be passed to make a command hide behind the admin guard.

### DIFF
--- a/Assets/Source/Input/Voice/IVoiceCommandManager.cs
+++ b/Assets/Source/Input/Voice/IVoiceCommandManager.cs
@@ -13,7 +13,16 @@ namespace CreateAR.EnkluPlayer
         /// <param name="keyword">The keyword to look for.</param>
         /// <param name="delegate">The delegate that will be called.</param>
         /// <returns></returns>
-        bool Register(string keyword, Action<string> @delegate, bool adminCommand = false);
+        bool Register(string keyword, Action<string> @delegate);
+
+        /// <summary>
+        /// Registers a keyword for detection.
+        /// Admin keywords are guarded by the "admin" phrase.
+        /// </summary>
+        /// <param name="keyword">The keyword to look for.</param>
+        /// <param name="delegate">The delegate that will be called.</param>
+        /// <returns></returns>
+        bool RegisterAdmin(string keyword, Action<string> @delegate);
 
         /// <summary>
         /// Unregisters a keyword.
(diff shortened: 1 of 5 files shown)


Assets/Source/Input/Voice/IVoiceCommandManager.cs
```diff
@@ -13,7 +13,7 @@ public interface IVoiceCommandManager
         /// <param name="keyword">The keyword to look for.</param>
         /// <param name="delegate">The delegate that will be called.</param>
         /// <returns></returns>
-        bool Register(string keyword, Action<string> @delegate);
+        bool Register(string keyword, Action<string> @delegate, bool adminCommand = false);
```
thegoldenmule: Rather than a function with an extra bool (which is hard to understand), I'd recommend two functions:

`Register`
`RegisterAdmin`

Assets/Source/Input/Voice/VoiceCommandManager.cs
```diff
@@ -12,6 +12,12 @@ namespace CreateAR.EnkluPlayer
     /// </summary>
     public class VoiceCommandManager : IVoiceCommandManager
     {
+        private struct PhraseConfig
```
thegoldenmule: Comments on this guy.

Assets/Source/Input/Voice/VoiceCommandManager.cs
```diff
@@ -57,21 +73,27 @@ public class VoiceCommandManager : IVoiceCommandManager
...
-        public bool Register(string keyword, Action<string> callback)
+        public bool Register(string keyword, Action<string> callback, bool adminCommand = false)
         {
             if (_actions.ContainsKey(keyword))
             {
                 return false;
             }
 
-            _actions[keyword] = callback;
+            _actions[keyword] = new PhraseConfig()
```
thegoldenmule: Unneeded `()`s-- do you have resharper yet?

jsundra: I do! Did you send me your resharper config yet? :P

Assets/Source/Input/Voice/IVoiceCommandManager.cs
jsundra: I like it! Don't really see a huge pro for it being a single method, I'll split it out.

PR #191, opened 2018-10-22
Bug/schema jsapi

jsundra:
Found a bug where using the SchemaJsApi `watch` functionality would throw null refs when invoking callbacks.

Changed the `thisArg` param back to a JsValue wrapped version of `this`. I elected to make it on the fly and not cache it by engine, since scripts constantly being edited/resaving/rerunning would come with a different engine and cause a memory leak. Eventually when we have one shared Engine instance, this can be cached easily.

Also updated the tests to not be runtime tests, so that they're more consistent with the existing jsapi tests and can be quickly run.

Misc: Fixed the Quat test that broke sometime last week.


Assets/Editor/Test/Scripting/ElementSchemaJsApi_Tests.cs
```diff
@@ -0,0 +1,112 @@
...
+            _element = new ElementJs(null, null, new Element());
+            _engine.SetValue("this", _element);
+        }
+
+        [Test]
+        public void GetSetNumber()
+        {
+            var output = _engine.Run(@"
+                this.this.schema.setNumber('foo', 12);
+                this.this.schema.getNumber('foo');
```
thegoldenmule: `this.this`?

jsundra:
Oh right, I forgot to mention that in the PR. Short answer, no clue?! I had to do it in an earlier test so copied that nuance over. I looked into it really briefly and nothing jumped out. By calling SetValue, jint is wrapping it in a JsValue just like the `this` binding is set in the runtime code.

It really looks dumb though, debating diving into it more or moving onto something else and ignoring the OCD cringe.

Assets/Editor/Test/Scripting/ElementSchemaJsApi_Tests.cs
```diff
@@ -0,0 +1,112 @@
...
+        [SetUp]
+        public void Setup()
+        {
+            _engine = new Engine(options => {
+                options.CatchClrExceptions(exception => { throw exception; });
+                options.AllowClr();
+            });
+
+            _element = new ElementJs(null, null, new Element());
+            _engine.SetValue("this", _element);
```
thegoldenmule: Oh, I see, you're setting `this` here. We should change to a different name, like `element`.

jsundra: OHH. Right and the difference between the test runner and the runtime usage is that the magic functions (enter/update/exit) are bound with their context before being invoked. Makes sense why the silliness was needed. Changing it!

PR #193, opened 2018-10-26
Anchoring Bugs

jsundra:
Fixes a few bugs with anchoring:

- Anchors are unable to be selected in the web editor.
`WorldAnchorWIdget` was being created with its `locked` schema set to true by default.

- Anchors made in the web editor wouldn't anchor in the correct position
New anchors were being placed immediately based on their schema & exported without first being repositioned. Adding the `none` anchor state to the relative positioning check fixed this.

- Anchors made in Hololens aren't creating properly
Similar to the web editor issue, Anchors weren't exporting with the correct transform data. Anchors now always re-position relatively before exporting.

- Multiple anchors from the web editor couldn't export
This one seemed to come for free. Tried it out a few times and multiple anchors all always exported.. Some anchoring code did change since we ran into that last, or maybe there is an issue where if one fails to save (poor wifi) the rest would fail.

To allow anchors to re-position themselves relatively, I added some support to the `IPrimaryAnchorManager`. It isn't perfect, and when more anchoring changes come in the next few weeks I want to revisit it. Right now anchors can be triggered for exporting or have their schema updated from a variety of different sources, it'd be nice to centralize that a bit more. I had to add an ugly hack to get around a circular dependency I created. I thought about adding `IElementFactory` to `VineImporter` and having that be responsible for creating Elements directly and not just their Descriptions, but that would have been a change that touched a lot of files and not something I wanted to do for a quick fix or without more thought.

### DIFF
--- a/Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
+++ b/Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
@@ -193,6 +193,8 @@ namespace CreateAR.EnkluPlayer
         private int _pollUnlocated;
         private int _pollLocated;
 
+        private Matrix4x4 _lastTransformMatrix;
+
         /// <inheritdoc />
         public WorldAnchorWidget.WorldAnchorStatus Status
         {
@@ -338,6 +340,11 @@ namespace CreateAR.EnkluPlayer
             {
                 _surfaces[id] = new SurfaceRecord(filter.gameObject);
             }
+        }
+
+        public void PositionRelatively(WorldAnchorWidget anchor)
+        {
+            PositionAnchorRelative(anchor, _lastTransformMatrix);
         }
 
         /// <summary>
@@ -615,7 +622,7 @@ namespace CreateAR.EnkluPlayer
         private void UpdateRelativePositioning()
         {
             var located = FirstLocatedAnchor();
-            var T = Matrix4x4.identity;
+            _lastTransformMatrix = Matrix4x4.identity;
 
             // calculate inverse transformation
             if (null != located)
@@ -630,21 +637,18 @@ namespace CreateAR.EnkluPlayer
                 var B = Matrix4x4.TRS(locatedPos, locatedRot, Vector3.one);
 
                 // T * A = B
-                T = B * Matrix4x4.Inverse(A);
+                _lastTransformMatrix = B * Matrix4x4.Inverse(A);
             }
 
             // place non-located anchors relative to the located anchor
             for (int i = 0, len = _anchors.Count; i < len; i++)
             {
                 var anchor = _anchors[i];
-                if (anchor.Status == WorldAnchorWidget.WorldAnchorStatus.IsError
-                    || anchor.Status == WorldAnchorWidget.WorldAnchorStatus.IsImporting
-                    || anchor.Status == WorldAnchorWidget.WorldAnchorStatus.IsLoading
-                    || anchor.Status == WorldAnchorWidget.WorldAnchorStatus.IsReadyNotLocated)
+                if (ShouldPositionRelatively(anchor))
                 {
                     if (null != located)
                     {
-                        PositionAnchorRelative(anchor, T);
+                        PositionAnchorRelative(anchor, _lastTransformMatrix);
                     }
                     else
                     {
@@ -654,6 +658,20 @@ namespace CreateAR.EnkluPlayer
                     }
                 }
             }
+        }
+
+        /// <summary>
+        /// Determines if an anchor should be manually repositioned or not.
+        /// </summary>
+        /// <param name="anchor">The anchor to check.</param>
+        /// <returns></returns>
+        private bool ShouldPositionRelatively(WorldAnchorWidget anchor)
+        {
+            return anchor.Status == WorldAnchorWidget.WorldAnchorStatus.None
+                   || anchor.Status == WorldAnchorWidget.WorldAnchorStatus.IsError
+                   || anchor.Status == WorldAnchorWidget.WorldAnchorStatus.IsImporting
+                   || anchor.Status == WorldAnchorWidget.WorldAnchorStatus.IsLoading
+                   || anchor.Status == WorldAnchorWidget.WorldAnchorStatus.IsReadyNotLocated;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


Assets/Source/Player/IUX/Widget/WorldAnchor/WorldAnchorWidget.cs
```diff
@@ -240,10 +250,10 @@ protected override void LoadInternalBeforeChildren()
             _versionProp = Schema.GetOwn("version", -1);
             _versionProp.OnChanged += Version_OnChanged;
 
-            _lockedProp = Schema.GetOwn("locked", true);
+            _lockedProp = Schema.GetOwn("locked", false);
             _lockedProp.OnChanged += Locked_OnChanged;
             
-            var autoexport = Schema.GetOwn("autoexport", false).Value;
+            var autoexport = Schema.GetOwn("autoexport", true).Value;
```
jsundra: I wasn't able to test this one line change yesterday. I got to a point where everything worked except placing new anchors in the Hololens, with that leading to an exception. I thought this was the issue, where Anchors weren't having the right code path executed when first being created, but my Hololens started acting finnicky (complaining about files in use when deploying) and I didn't have time to sort it out. 

Looking at it now in the morning, I don't think it's perfect either. The issue would be with the schema values not being re-positioned correctly, not with the GameObject's transform being wrong like when they come in from the web.

Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
```diff
@@ -193,6 +193,8 @@ public SurfaceRecord(GameObject gameObject)
         private int _pollUnlocated;
         private int _pollLocated;
 
+        private Matrix4x4 _lastTransformMatrix;
```
thegoldenmule: I'm a little confused about this guy. Why is this necessary? Seems like something that can go stale/introduce bugs very easily. Looks like it's motivated by `PositionRelatively`, but that could just use a reference to the primary anchor internally.

Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
```diff
@@ -924,7 +943,11 @@ private void Messages_OnAutoExport(object @object)
                 Log.Info(this, "Primary is located. Positioning AutoExport anchor.");
 
                 // add
-                _anchors.Add(anchor);
+                if (!_anchors.Contains(anchor))
+                {
+                    _anchors.Add(anchor);
+                }
```
thegoldenmule: Were anchors getting double added? If so, seems like a deeper bug-- like this method should not be getting called multiple times.

Assets/Source/Main.cs
```diff
@@ -65,6 +65,15 @@ public static AppJsApi NewAppJsApi(IElementJsCache jsCache)
...
 	    }
 
+		/// <summary>
+		/// TODO: REMOVE THIS UNHOLY ABOMINATION. This is here because of a circular dependency between IElementFactory < == > IPrimaryAnchorManager
+		/// </summary>
+		/// <returns></returns>
+		public static IPrimaryAnchorManager PrimaryAnchorManager()
+		{
+			return _binder.GetInstance<IPrimaryAnchorManager>();
+		}
```
thegoldenmule: I have set a poor example. I don't think we should allow stuff like this into `master`. Let's try to fix that circular dependency.

Assets/Source/Player/IUX/Widget/WorldAnchor/WorldAnchorWidget.cs
```diff
@@ -178,6 +185,9 @@ public void Reload()
         /// <param name="txns">Object to make txns with.</param>
         public void Export(string appId, string sceneId, IElementTxnManager txns)
         {
+            // Update this Anchor's position relative to a located Anchor
+            _anchorManager.PositionRelatively(this);
```
thegoldenmule: I'm suspicious of this. Are we sure this is what we want? Exporting works fine when you export a primary anchor and other anchors in the same session, i.e. they were all exported from the same world space. Seems like this needs to export the position/rotation relative to that space. Maybe that's what this is doing, I'm confused.

Assets/Source/Player/IUX/Widget/WorldAnchor/WorldAnchorWidget.cs
thegoldenmule: Let's retest then.

Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
jsundra: Right now it's only ever double called. When anchors are created from the web, they're initially added in the scene graph lookup, then their `ANCHOR_AUTOEXPORT` message is dispatched. It looks like it was intended to be used for when any anchor wanted to export, to use that flow, but currently anchor exporting is handled in a variety of places from `PrimaryAnchorManager` to the different `AnchorDesignState`s. "Consolidate anchor export flows" is pretty high on my notes for an anchoring refactor.

Assets/Source/Player/IUX/Widget/WorldAnchor/WorldAnchorWidget.cs
jsundra: For the case of web anchors, that's what this does. For hololens, that's where it's running into the bug I mentioned. Going to change things a bit to not make this dependency explicitly called every time, but also to make it stronger than just being a forgettable call on one of the different ways anchors export currently.

Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
jsundra: I cached the last value with the idea that in the near-future, we might be able to do a comparison between the current offset matrix and the last, and skip all the looping/maths/transform interactions on other anchors if they were the same/similar. I haven't looked into the offset yet though to see if the world positioning changes slightly every frame or if it's stable enough to potentially get a small gain from this.

But yeah, agreed that in the current state it isn't robust and could return a stale transformation if tracking is lost.

Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
thegoldenmule: Ahhh, good to know.

PR #195, opened 2018-11-05
Feature/script runner

jsundra:
Super rough WIP of the new `ScriptService` & `ScriptRunner`. Currently functionality is supposed to mimic the existing setup/running of scripts via `ContentWidget`->`ScriptCollectionRunner`. Scripts are still expected to run in edit mode, and have no hierarchy based script execution order yet. Still have to go and cleanup things, add docstrings on the new systems, etc.

## Changes
- `IScriptFactory` now takes `Widget`s instead of `GameObject`s and `Element`s.
- `IAppSceneManager` has a prop for getting its initialized token outside of the Initialize invoke.
- Added a base `Script` that `VineScript` and `BehaviorScript` extend. (Maybe make this `EnkluScript` and make the existing one more data-centric). This makes `EnkluScriptElementBehavior` have potential to parse async, though it currently does everything synchronously still.

### ScriptRunner
- "ScriptManager" from the scripting planning doc. Similar to ScriptCollectionRunner in responsibilities, but manages scripts for multiple widgets instead of being created 1:1.
- Loads scripts for `Widget`s, configures them, and handles invoking their lifecycle functions.
- Waits for all Vines on all widgets before handling any Behaviors.
TODO:
- Handle configuration errors (they currently will still get `update` invokes even if they failed).
- Handle script src changes.
- Handle exiting/stopping scripts.

### ScriptService
- Waits for the experience to finish initializing, then iterates the scene & listens for new widget creation.
TODO: 
- Handle Element destruction / script unloading
- Handle experience changing


Assets/Editor/Test/Scripting/ScriptService_Tests.cs
```diff
@@ -0,0 +1,118 @@
...
+            var vineComponent = _scriptFactory.GetVine(_vines[0]);
+            var behaviorComponent = _scriptFactory.GetBehavior(_behaviors[0]);
+            
+            Assert.AreEqual(0, vineComponent.EnterInvoked);
+            Assert.AreEqual(0, behaviorComponent.EnterInvoked);
+            
+            vineComponent.FinishConfigure();
+            
+            Assert.AreEqual(1, vineComponent.EnterInvoked);
+            Assert.AreEqual(1, behaviorComponent.EnterInvoked);
```
thegoldenmule: I'm a tad confused on this interface. So `EnterInvoked` is a count of how many times `enter` has been called? That seems odd.

Assets/Source/Player/Element/Txns/AppSceneManager.cs
```diff
@@ -28,6 +33,12 @@ public string[] All
                 return _scenes.Keys.ToArray();
             }
         }
+        
+        /// <inheritdoc />
+        public IAsyncToken<Void> OnInitialized
```
thegoldenmule: This is not a good solution. `IAsyncToken` is for completion values that are immutable. `AppSceneManager`, however, is a persistent, reusable object. `Initialize` and `Uninitialize` can be called many times, but the token can only be resolved once. An event should be used instead. If necessary, an event and a `bool`.

Assets/Source/Player/IUX/Content/ScriptCollectionRunner.cs
```diff
@@ -231,7 +233,7 @@ private IAsyncToken<Void> RunVine(EnkluScript script)
         {
             Log.Info(this, "Run vine({0}) : {1}", script.Data, script.Source);
 
-            var component = _scriptFactory.Vine(_root, _element, script);
+            var component = _scriptFactory.Vine(_element as Widget, script);
```
thegoldenmule: If you know it's a `Widget`, then direct cast. If it's allowed to not be a `Widget`, you should be checking for null. This doesn't seem safe.

Assets/Source/Player/Scripting/Script.cs
```diff
@@ -0,0 +1,28 @@
+using CreateAR.Commons.Unity.Async;
+using UnityEngine;
+
+namespace CreateAR.EnkluPlayer.Scripting
+{
+    public abstract class Script : MonoBehaviour
```
thegoldenmule: Not a big fan of a pure abstract class here. An abstract class is useful when it actually implements some common behavior-- but this is just acting as an interface (except it requires MonoBehavior), so it should be changed into an interface.

Assets/Source/Player/Scripting/ScriptFactory.cs
```diff
@@ -35,24 +36,33 @@ public class ScriptFactory : IScriptFactory
...
         
         /// <inheritdoc />
-        public VineScript Vine(GameObject root, Element element, EnkluScript script)
+        public VineScript Vine(Widget widget, EnkluScript script)
         {
-            var component = root.AddComponent<VineMonoBehaviour>();
-            component.Initialize(element, script, _vineImporter, _elementFactory);
+            if (widget == null)
+            {
+                throw new Exception("Vine attached to non-widget?!");
```
thegoldenmule: This exception doesn't match the case it's handling. Widget being null does not mean a Vine is attached to a non-widget. This should be more specific.

Assets/Source/Player/Scripting/ScriptFactory.cs
```diff
@@ -35,24 +36,33 @@ public class ScriptFactory : IScriptFactory
...
             UnityScriptingHost host, 
-            EnkluScript script, 
-            Element element)
+            EnkluScript script)
         {
-            var component = root.AddComponent<EnkluScriptElementBehavior>();
-            component.Initialize(jsCache, _elementJsFactory, host, script, element);
+            if (widget == null)
+            {
+                throw new Exception("Vine attached to non-widget?!");
```
thegoldenmule: Same here, plus it's not a Vine.

Assets/Source/Player/Scripting/ScriptRunner.cs
```diff
@@ -0,0 +1,315 @@
...
+            {
+                throw new Exception("Widget already added.");
+            }
+            
+            _widgets.Add(widget);
+            _scriptLookup.Add(widget, new List<EnkluScript>());
+            _vines.Add(widget, new List<VineScript>());
+            _behaviors.Add(widget, new List<BehaviorScript>());
+            _engines.Add(widget, CreateBehaviorHost(widget));
+            _setupStates.Add(widget, SetupState.None);
```
thegoldenmule: Lots of parallel lookups are reaaaaaal easy to mess up. Consider aggregating all this into an object. Then it's `widget` => `WidgetRecord`, which then has a bunch of readonly `List`s that are all together on one object.

Assets/Source/Player/Scripting/ScriptRunner.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            try
+            {
+                _scriptLoading.Add(CreateScripts(widget));
+            }
+            catch (Exception e)
```
thegoldenmule: I'm not sure an exception is the best way to handle this. `CreateScripts` is doing a _lot_ of state changes, and none of them can be rolled back if you just throw an exception in the middle.

Assets/Source/Player/Scripting/ScriptRunner.cs
```diff
@@ -0,0 +1,315 @@
...
+        {
+            // Vines currently have no Update usage, so skip for now.
+
+            for (int i = 0, iLen = _widgets.Count; i < iLen; i++)
+            {
+                var behaviors = _behaviors[_widgets[i]];
+
+                for (int j = 0, jLen = behaviors.Count; j < jLen; j++)
+                {
+                    behaviors[j].FrameUpdate();
```
thegoldenmule: What if a behavior adds or removes a behavior in an update loop?

Assets/Source/Player/Scripting/ScriptRunner.cs
```diff
@@ -0,0 +1,315 @@
...
+using System.Linq;
+using CreateAR.Commons.Unity.Async;
+using CreateAR.Commons.Unity.Logging;
+using CreateAR.EnkluPlayer.IUX;
+using Newtonsoft.Json.Linq;
+using Void = CreateAR.Commons.Unity.Async.Void;
+
+namespace CreateAR.EnkluPlayer.Scripting
+{
+    public class ScriptRunner
```
thegoldenmule: I'll be curious to see this more worked out. My initial impression is that it's very confusing, however I do admit it's a confusing problem.

Assets/Source/Player/Scripting/ScriptService.cs
```diff
@@ -0,0 +1,72 @@
...
+            
+            _scriptRunner = new ScriptRunner(
+                scriptManager, scriptFactory, requireResolver, elementJsCache);
+        }
+
+        public override void Start()
+        {
+            base.Start();
+
+            _sceneManager.OnInitialized.OnSuccess((_) =>
```
thegoldenmule: As I pointed out in the scene manager file, this is dangerous-- you want an event instead.

Assets/Source/Player/Scripting/ScriptService.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                for (int i = 0, len = _elementManager.All.Count; i < len; i++)
+                {
+                    OnElementCreated(_elementManager.All[i]);
```
thegoldenmule: Minor: for iteration over properties, cache the property locally. There is no JITer to hotwire this.

Assets/Editor/Test/Scripting/ScriptRunner_Tests.cs
```diff
@@ -0,0 +1,185 @@
...
+        [Test]
+        public void Vine()
+        {
+            var widget = CreateWidget(_vines[0]);
+            
+            _scriptRunner.AddWidget(widget);
+            _scriptRunner.ParseAll();
+            
+            // Vines are async, check that it hasn't finished.
+            Assert.AreEqual(ScriptRunner.SetupState.Parsing, _scriptRunner.GetSetupState(widget));
```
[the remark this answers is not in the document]
jsundra: The test implementation of `VineScript`s don't resolve their configure token until `FinishConfigure` is called. So there's no chance tests will have a false positive due to race conditions.

Assets/Editor/Test/Scripting/ScriptRunner_Tests.cs
```diff
@@ -0,0 +1,185 @@
...
+        [Test]
+        public void Behavior()
+        {
+            var widget = CreateWidget(_behaviors[0]);
+            
+            _scriptRunner.AddWidget(widget);
+            _scriptRunner.ParseAll();
+            
+            // Behaviors load synchronously. Check that the script has loaded & invoked.
+            Assert.AreEqual(ScriptRunner.SetupState.Done, _scriptRunner.GetSetupState(widget));
```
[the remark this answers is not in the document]
jsundra: Only for now! To mimic `ScriptCollectionRunner`, everything parses -> executes. Once all of the current functionality is implemented, the immediate execution will be removed so that scripts will be able to run only in play mode.

Assets/Editor/Test/Scripting/ScriptService_Tests.cs
jsundra: Yeah, it might not be something super standard but a pattern I usually use in tests. When checking function/callback invokes, I don't like to rely on just setting a bool since that doesn't safeguard the chance that something can be invoked multiple times.

Assets/Source/Player/Element/Txns/AppSceneManager.cs
jsundra: Ahh good point. I'll revisit this and make it more robust for experience reloads/changes. Thanks!

Assets/Source/Player/IUX/Content/ScriptCollectionRunner.cs
jsundra: Yeah definitely. This was just thrown in to make everything compile/work in a controlled environment when switching back and forth between the old/new script execution systems. I was planning to fully remove `ScriptCollectionRunner` as part of the feature branch.

Assets/Source/Player/Scripting/Script.cs
jsundra: Agreed. I used an abstract class over an interface in an early PR for some necessity, I think when vines were relying on `InjectableMonobehaviour`. So I should be able to go back to using interfaces for this now. Good catch!

Assets/Source/Player/Scripting/ScriptRunner.cs
jsundra: Ah right, I didn't think about the use case where scripts modify the script src schema. It'd be pretty hard to add a script without knowing its id, but yeah something to guard against.

thegoldenmule: I was thinking more when we add an interface to add a script at runtime.

Assets/Editor/Test/Scripting/ScriptService_Tests.cs
thegoldenmule: Yes, but it doesn't look like that is a test class. That's in runtime code.

Assets/Source/Player/Scripting/ScriptService.cs
jsundra: A bit confused here... are you suggesting doing something like this:
```
for (int i = 0, len = _elementManager.All.Count; i < len; i++)
{
    var element = _elementManager.All[i];
    OnElementCreated(element);
...
```
Probably misunderstanding since I don't see that big of a difference. Can you explain the JIT hotwiring you're referring to?

thegoldenmule:
```
var all = _elementManager.All;
for { ... }
```

A compiler like gcc would probably replace this with direct lookups, rather than a property lookup and then a lookup (i.e. automatically replace the extra indirection). In C# there is an AOT compiler (like gcc) and a JIT compiler-- the former generally does less because the latter can do runtime optimizations. This speeds up development time because compile times are much lower. Because we're ripping out the JITer when we compile to iOS or using IL2CPP (where only the AOT compiler is used), we only get the less optimized AOT compiler output. I have not run tests with the latest Mono compiler Unity is using, but in previous projects I've had to get in the habbit using for instead of foreach, caching length, and caching properties like this in loops.

jsundra:
Going to close/reopen this when it's ready again!

PR #196, opened 2018-11-06
Asset Versioning

thegoldenmule:
Needs more testing, but it's good to get ahead on code review.

* `Asset` now does much less. Each `Asset` is version locked and does not auto-update. This actually makes it much more testable and removes the need for many integration tests.
* Moved update events into `AssetManifest`.
* Update logic has been moved into the `AssetAssembler`.
* Backward compatible with old asset uris.


Assets/Editor/Test/Asset/AssetManifest_Tests.cs
```diff
@@ -50,30 +50,30 @@ public void Info()
         [Test]
         public void Get()
         {
-            Assert.AreSame("a", _manifest.Asset("a").Data.Guid);
+            Assert.AreSame("a", _manifest.Asset("a", -1).Data.Guid);
```
thegoldenmule: I should probably make -1 the default.

Assets/Source/Application/Services/AssetUpdateService.cs
```diff
@@ -186,6 +202,16 @@ private void FormatAssetData(AssetData asset)
                 asset.Uri = asset.Uri.Replace(
                     "{{platform}}",
                     GetBuildTarget());
+
+                // v3
+                var pieces = asset.Uri.Split('.');
```
thegoldenmule: Unsafe way of determining version, really. Not sure what a good alternative is atm.

Assets/Editor/Test/Asset/AssetManifest_Tests.cs
```diff
@@ -182,16 +164,105 @@ public void UpdateEvent()
...
+            });
+            unwatch();
+
+            _manifest.Update(info);
+
+            Assert.IsFalse(eventCalled);
+        }
+
+        [Test]
+        public void WatchGoodUpdate()
```
jsundra: This test name is kind of confusing. Wasn't clear at first glance that it was supposed to watch another element to make sure the callback didn't invoke. Of course, I don't have a good/short suggestion though :/

Assets/Editor/Test/Asset/Asset_Tests.cs
```diff
@@ -125,258 +126,11 @@ public void LoadComponent()
...
-            Assert.AreEqual(0, watches);
-
-            _reference.AutoReload = true;
-
-            Assert.AreEqual(1, watches);
-
-            _reference.Update(DataUpdate);
-
-            Assert.AreEqual(2, watches);
+            Assert.IsFalse(string.IsNullOrEmpty(_errorReference.Error));
```
jsundra: Shouldn't this be true if the test is "LoadError"? Maybe it'll make more sense as I scroll down more.

Assets/Source/Application/Services/AssetUpdateService.cs
jsundra: Only suggestion I'd have is adding in a UriVersion to the data structure / db. The v1/v2 logic can stay if there's no UriVersion, and rely on that for v3+? Might not be the quickest thing to add though, I haven't looked into asset ingest at all yet.

Assets/Source/Player/Asset/StandardAssetLoader.cs
```diff
@@ -248,9 +255,21 @@ private IEnumerator ProcessQueue()
...
+
+            // try to insert version
+            var pieces = data.Uri.Split('.');
+            if (pieces.Length == 3)
+            {
+                uri = string.Format("{0}.{1}.{2}",
+                    pieces[0],
+                    version,
+                    pieces[2]);
+            }
```
jsundra: Maybe this logic can be split off into a Versioning helper? Even just a static class/method. I know the asset/thumb URIs are different for v1/v2, but this v3 logic seems the same in the 3 places it's used in this PR.

enkluplayer.sln.DotSettings
```diff
@@ -0,0 +1,5 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=Constants/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AA_BB" /&gt;</s:String>
+	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=TypesAndNamespaces/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AaBb"&gt;&lt;ExtraRule Prefix="" Suffix="_Tests" Style="AaBb_AaBb" /&gt;&lt;/Policy&gt;</s:String>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=coroutines/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Enklu/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
```
jsundra: 🏆

Assets/Source/Application/Services/AssetUpdateService.cs
thegoldenmule: I'm going to add a preceding `v3:`. This will allow future changes to incorporate a version too.

Assets/Editor/Test/Asset/Asset_Tests.cs
thegoldenmule: We want it to be false that the error is empty. Which is a dumb way of saying we want the error to be non-empty.

docs/application.md
```diff
@@ -0,0 +1,19 @@
...
+
+##### Application
+
+An instance of `Application` is created in `Main.cs`. We use a dependency injection framework to do this-- the details are not covered here. This object is incredibly simple. It contains an `IApplicationServiceManager` that it forwards function calls to. These functions are the essential lifecycle functions of a Unity application.
+
+Inside the `ApplicationServiceManager`, there is a collection of `ApplicationService` objects. These are objects that run for the lifetime of the application and contain the meat of the application logic.
+
+##### ApplicationService
+
+These objects are designed to be created and once and used once in a simple flow: `Start` -> `Update` -> `Stop`. Internally, the `ApplicationService` also provides a succinct wrapper for the [messaging system](http://docs.enklu.com/commons-unity/index.html#messaging) that automatically unsubscribes from messages on teardown.
```
jsundra: Typo: ...designed to be created and once and used once...

docs/application.stateservice.md
```diff
@@ -0,0 +1,41 @@
...
+    void Update(float dt);
+
+    /// <summary>
+    /// Called when the state is transitioned out.
+    /// </summary>
+    void Exit();
+}
+```
+
+An FSM then, is equally simple, exposing a single `ChangeState<T>` method that switching between states. All major states of the application are implemented using this simple interface. Our states also do not contain logic about what states to switch into. They simply dispatch events about what they are doing.
```
jsundra: Typo: ...method that switching between...

Assets/Source/Application/Services/AssetUrlHelper.cs
```diff
@@ -0,0 +1,167 @@
+using System;
+using CreateAR.Commons.Unity.Logging;
+using UnityEngine;
+
+namespace CreateAR.EnkluPlayer
+{
+    /// <summary>
+    /// Provides helper methods for dealing with assets.
+    /// </summary>
+    public static class AssetUrlHelper
```
Pixelwelder: Just a general note (that may be totally off):

It seems weird to see utility methods that require more information than they use. These three methods format a string into a different string. However, they require that string to be wrapped in an Asset. I get that they're only _used_ for Assets, but philosophically it's weird to me.

Anyway, approved.

thegoldenmule: Part of the whole issue here is that different versions of asset URLs require different parts of asset data. V2 urls, for instance, don't require anything. The desire here is to define a function that maps an `Asset` to a `string`. Mapping a `string, string, number` tuple to a `string` is not straightforward and the caller would already have to know what version of URL we were using.

Assets/Source/Player/Asset/StandardAssetLoader.cs
thegoldenmule: Yes, moved it all into `AssetUrlHelper`.

PR #203, opened 2018-11-09
Remove black background from Gesture.png

jsundra:
Forgot to do this when initially uploading it. Removes the black background so it doesn't awkward block content behind it


jsundra:
Requested a review out of habit. Just going to merge this since the diff doesn't even show anything

PR #207, opened 2018-11-16
Feature/world transforms

jsundra:
## Maths
- Added Inverse for Quats, global position/rotation/scale for transforms as well as `transformPoint`.
- Added tests for the new functionality.

## Misc
- Renamed `ContentElementJs` -> `ContentWidgetJs`.
- Fixed up some line endings.


Assets/Source/Player/Scripting/Interface/Elements/ElementTransformJsApi.cs
```diff
@@ -108,5 +135,20 @@ public void lookAt(Vec3 direction)
...
+        {
+            // Scale the incoming vec.
+            var cachedScale = worldScale;
+            var rtn = new Vec3(src.x * cachedScale.x, src.y * cachedScale.y, src.z * cachedScale.z);
+            
+            // Rotate it...
+            rtn = Quat.Mult(worldRotation, rtn);
+            
+            // Reposition it!
+            rtn = rtn + worldPosition;
```
thegoldenmule: I think you want to use matrix multiplication here, not separate scale, rotate, translate. Look at `Matrix4x4`.

thegoldenmule: @jsundra once this is in, I think we can merge it in. While I disagree with the approach, I think the interface is fine-- meaning we can add our own world positioning system underneath at a later date _when_ it becomes a problem. ;)

PR #209, opened 2018-11-16
Bugfix/edit mode

thegoldenmule:


thegoldenmule:
Fixes issue where sometimes the last loaded scene was not saved to user preferences.

PR #211, opened 2018-11-27
Bugfix/graphite

thegoldenmule:
Seeing lots of HostedGraphite errors.

* These should be warnings, not errors.
* We should retry failed metrics.
* We should not attempt to write to DataWriter while it is flushing (I believe this was the cause of some errors).
* TODOs for adding fallbacks.
* Split Framework off from Uwp implementation.


Assets/Source/Util/Analytics/HostedGraphite/FrameworkHostedGraphiteMetricsTarget.cs
```diff
@@ -0,0 +1,112 @@
...
+        public void Send(string key, float value)
+        {
+            var message = string.Format("{0}.{1} {2}\n", _appKey, key,
+                value);
+            var bytes = Encoding.ASCII.GetBytes(message);
+
+            if (null != _socket)
+            {
+                _socket.SendTo(bytes, _endpoint);
+            }
```
jsundra: Minor, but all of this could fall under the socket check

Assets/Source/Util/Analytics/HostedGraphite/UwpHostedGraphiteMetricsTarget.cs
```diff
@@ -0,0 +1,251 @@
...
+            /// </summary>
+            private double _timestamp;
+
+            /// <summary>
+            /// Tracks number of tries.
+            /// </summary>
+            private int _tries;
+
+            /// <summary>
+            /// Trie iff retried enough times.
```
jsundra: Tipo!

Assets/Source/Util/Analytics/HostedGraphite/FrameworkHostedGraphiteMetricsTarget.cs
thegoldenmule: Done

Assets/Source/Util/Analytics/HostedGraphite/UwpHostedGraphiteMetricsTarget.cs
thegoldenmule: Done.

PR #214, opened 2018-11-29
Tween library

thegoldenmule:
Creating the PR now. Definitely a WIP.


thegoldenmule:
@jsundra Opened this for visibility. Going to stub out some other tween types (vec3, col4) and then start on the JS API.

I wanted this to be a separate C# system, not just a JS interface. Also, I wanted a data object (`TweenData`) so that we can serialize these suckers if we want.

Assets/Source/Player/Scripting/Interface/TweenJs.cs
```diff
@@ -0,0 +1,238 @@
...
+                        var callback = _onStart[i];
+                        _tween.OnStart += () => callback(context, new JsValue[0]);
+                    }
+
+                    // add completes
+                    for (int i = 0, len = _onComplete.Count; i < len; i++)
+                    {
+                        var callback = _onComplete[i];
+                        _tween.OnComplete += () => callback(context, new JsValue[0]);
+                    }
```
jsundra: Minor, but these lists could be nulled out after the callbacks are all added to the tween.

Assets/Source/Util/Tween/Tween.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// Called with an evaluated parameter.
+        /// </summary>
+        /// <param name="time">The time.</param>
+        protected abstract void Update(float time);
+    }
+
+    /// <summary>
+    /// Tween for a float.
+    /// </summary>
+    public class FloatTween : Tween
```
jsundra: Might be good to break all the implementations into their own files as we add support for various different data types.

thegoldenmule: For these small classes, I'd rather keep them all in one place.

Assets/Source/Player/Scripting/Interface/TweenJs.cs
thegoldenmule: :+1:

thegoldenmule:
Let's leave it off for now and see if we need it. We don't want to encourage an update loop!

PR #215, opened 2018-12-03
QR Code Hotfix

thegoldenmule:
* Clean up AV feed resources after reading QR code. This fixes voice commands.
* Log everything to a file.


Assets/Source/Application/Injection/EnkluPlayerModule.cs
```diff
@@ -424,7 +424,7 @@ private void SetupLogging(InjectionBinder binder, LogAppConfig config)
                     "Application.log"))
             {
                 // warning and up
-                Filter = LogLevel.Warning
+                Filter = LogLevel.Debug
```
jsundra: This should probably go back to `Warning`?

PR #217, opened 2018-12-03
Feature/schema jsapis

jsundra:
Updates to `AudioJsApi`, `MaterialJsApi`, and `AnimatorJsApi` to support values being backed by schema, so they can be used with the new tweening library.

The three JsApis check for schema first to load default values - otherwise they create schema entries off of the prefab's values. There are a few exceptions, `MaterialJsApi` has some notes for the next Unity upgrade. `AnimatorJsApi` lazily creates schema since only parameter names are known, not type.

All three have tests for loading with/without existing schema entries, and to ensure updating thru the scripting API or schema will reflect on the other. Since handling Materials, AudioSources, and Animators inside of tests is painful without a real material/audio/animator, etc, a lot of this PR ballooned into being just boilerplate for creating interfaces to wrap a lot of the Unity APIs so we can provide test implementations. @thegoldenmule let me know if you have other thoughts for this or don't want the interfaces to be as 1:1 with Unity.

### DIFF
--- a/Assets/Source/Player/IUX/Element/ElementSchema.cs
+++ b/Assets/Source/Player/IUX/Element/ElementSchema.cs
@@ -414,6 +414,11 @@ namespace CreateAR.EnkluPlayer.IUX
 
             return Default<T>();
         }
+
+        public ElementSchemaProp GetOwn(string name)
+        {
+            return Prop(name);
+        }
         
         /// <summary>
         /// Returns true iff the schema or parent schemas have a property with
--- a/Assets/Source/Player/IUX/Element/ElementSchema.cs
+++ b/Assets/Source/Player/IUX/Element/ElementSchema.cs
@@ -415,6 +415,11 @@ namespace CreateAR.EnkluPlayer.IUX
             return Default<T>();
         }
 
+        /// <summary>
+        /// Returns the prop if locally defined, or null otherwise.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
         public ElementSchemaProp GetOwn(string name)
         {
             return Prop(name);
--- a/Assets/Source/Player/IUX/Element/ElementSchema.cs
+++ b/Assets/Source/Player/IUX/Element/ElementSchema.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Text;
 
 namespace CreateAR.EnkluPlayer.IUX
@@ -414,16 +415,6 @@ namespace CreateAR.EnkluPlayer.IUX
 
             return Default<T>();
         }
-
-        /// <summary>
-        /// Returns the prop if locally defined, or null otherwise.
-        /// </summary>
-        /// <param name="name"></param>
-        /// <returns></returns>
-        public ElementSchemaProp GetOwn(string name)
-        {
-            return Prop(name);
-        }
         
         /// <summary>
         /// Returns true iff the schema or parent schemas have a property with
@@ -471,6 +462,16 @@ namespace CreateAR.EnkluPlayer.IUX
             }
 
             return false;
+        }
+        
+        /// <summary>
+        /// Returns a ReadOnlyCollection of all props specific to this Element.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        public ReadOnlyCollection<ElementSchemaProp> GetOwnProps()
+        {
+            return _props.AsReadOnly();
         }
 
         /// <summary>
--- a/Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
+++ b/Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using CreateAR.Commons.Unity.Logging;
 using CreateAR.EnkluPlayer.IUX;
 using UnityEngine;
@@ -152,7 +153,7 @@ namespace CreateAR.EnkluPlayer.Scripting
         /// </summary>
         /// <param name="name"></param>
         /// <returns></returns>
-        public int getInteger(string name)
+        public int getInt(string name)
         {
             return _animator.GetInt(name);
         }
@@ -162,7 +163,7 @@ namespace CreateAR.EnkluPlayer.Scripting
         /// </summary>
         /// <param name="name"></param>
         /// <param name="value"></param>
-        public void setInteger(string name, int value)
+        public void setInt(string name, int value)
         {
             var prop = GetSchemaProp(_propsInt, name);
             if (prop == null)
@@ -177,6 +178,30 @@ namespace CreateAR.EnkluPlayer.Scripting
             {
                 prop.Value = value;   
             }
+        }
+
+        /// <summary>
+        /// Gets the current value for a parameter.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        [Obsolete]
+        public int getInteger(string name)
+        {
+            Log.Warning(this, "animator.getInteger is deprecated. Use animator.getInt instead.");
+            return getInt(name);
+        }
+
+        /// <summary>
+        /// Sets the value for a parameter.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="value"></param>
+        [Obsolete]
+        public void setInteger(string name, int value)
+        {
+            Log.Warning(this, "animator.setInteger is deprecated. Use animator.setInt instead.");
+            setInt(name, value);
         }
 
         /// <summary>
--- a/Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
+++ b/Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
@@ -19,9 +19,9 @@ namespace CreateAR.EnkluPlayer.Scripting
         private readonly ElementSchema _schema;
         
         /// <summary>
-        /// Underlying Unity Animator to wrap.
+        /// Backing IAnimator.
         /// </summary>
-        private readonly Animator _animator;
+        private readonly IAnimator _animator;
 
         /// <summary>
         /// List of available parameter names.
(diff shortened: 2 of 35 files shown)


Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
```diff
@@ -31,15 +47,72 @@ public string[] parameterNames
...
+                        _animator.SetBool(_parameterNames[i], propBool.Value);
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Destructor.
+        /// </summary>
+        ~AnimatorJsApi()
```
thegoldenmule: Finalizers are useful for freeing unmanaged resources, but here I don't believe it is doing anything useful. We have no control over when this is called. We should only be using finalizers when implementing the `IDisposable` pattern. If there is a specific need for removing the handlers, then we should have a `Setup` and `Teardown` flow that adds/removes handlers.

Can you describe what bug you were seeing that caused you to add this?

Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
```diff
@@ -60,27 +133,75 @@ public bool getBool(string name)
...
+            Log.Warning(this, "animator.getInteger is deprecated. Use animator.getInt instead.");
+            return getInt(name);
         }
 
         /// <summary>
         /// Sets the value for a parameter.
         /// </summary>
         /// <param name="name"></param>
         /// <param name="value"></param>
+        [Obsolete]
```
thegoldenmule: Should be `[Obsolete("User animator.setInt instead.")]`.

Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
```diff
@@ -121,17 +254,82 @@ public string getCurrentClipName(int layer = 0)
...
-            var clips = _animator.GetCurrentAnimatorClipInfo(layer);
+            return _animator.IsClipPlaying(clipName, layer);
+        }
+        
+        /// <summary>
+        /// Converts a schema name to a parameter name.
+        /// </summary>
+        /// <param name="param"></param>
+        /// <returns></returns>
+        private string FromSchemaName(string schema)
```
thegoldenmule: `static`?

Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
```diff
@@ -121,17 +254,82 @@ public string getCurrentClipName(int layer = 0)
...
+        {
+            return schema.Substring(schema.IndexOf('.') + 1);
+        }
+
+        /// <summary>
+        /// Converts a parameter name to a schema name.
+        /// </summary>
+        /// <param name="param"></param>
+        /// <returns></returns>
+        private string ToSchemaName(string param)
```
thegoldenmule: `static`?

Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
```diff
@@ -121,17 +254,82 @@ public string getCurrentClipName(int layer = 0)
...
+            return "animator." + param;
+        }
+
+        /// <summary>
+        /// Update the underlying animator based on schema changes.
+        /// </summary>
+        /// <param name="prop"></param>
+        /// <param name="prev"></param>
+        /// <param name="new"></param>
+        private void OnSchemaFloatChange(ElementSchemaProp<float> prop, float prev, float @new)
```
thegoldenmule: To follow our convention, it should be `Prop_OnFloatChanged`, i.e. `[Source]_[Name]`.

Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
thegoldenmule: Also: not necessary here, but perhaps namespacing should be a first class part of schema somehow. Just thinking about it.

Assets/Source/Player/Scripting/Interface/Elements/AudioJsApi.cs
```diff
@@ -10,87 +12,124 @@ public class AudioJsApi
...
+                _volumeProp.Value = volume;
+            }
+
+            _volumeProp.OnChanged += OnVolumeChanged;
+        }
+
+        /// <summary>
+        /// Destructor.
+        /// </summary>
+        ~AudioJsApi()
```
thegoldenmule: Same here.

Assets/Source/Player/Scripting/Interface/Elements/AudioJsApi.cs
```diff
@@ -10,87 +12,124 @@ public class AudioJsApi
...
         /// <summary>
         /// Doppler level.
         /// </summary>
         public float dopplerLevel
         {
-            get { return _audio.dopplerLevel; }
-            set { _audio.dopplerLevel = value; }
+            get { return _audio.DopplerLevel; }
+            set { _audio.DopplerLevel = value; }
+        }
```
thegoldenmule: All the properties should be above the contstructor.

Assets/Source/Player/Scripting/Interface/Elements/MaterialJsApi.cs
```diff
@@ -1,45 +1,148 @@
...
+            _propsFloat = new List<ElementSchemaProp<float>>();
+            _propsInt = new List<ElementSchemaProp<int>>();
+            _propsVec3 = new List<ElementSchemaProp<Vec3>>();
+            _propsCol4 = new List<ElementSchemaProp<Col4>>();
+        }
+
+        /// <summary>
+        /// Destructor.
+        /// </summary>
+        ~MaterialJsApi()
```
thegoldenmule: And here.

Assets/Source/Player/Scripting/Interface/IAnimator.cs
```diff
@@ -0,0 +1,72 @@
+using UnityEngine;
+
+namespace CreateAR.EnkluPlayer.Scripting
+{
+    /// <summary>
+    /// Interface for an Animator.
+    /// </summary>
+    public interface IAnimator
```
thegoldenmule: I probably just would have had the tests instantiate a prefab with an animator (some of the other tests do this), but this is probably better.

Assets/Source/Player/IUX/Element/ElementSchema.cs
```diff
@@ -414,6 +414,16 @@ public ElementSchemaProp<T> GetOwn<T>(string name, T @default)
 
             return Default<T>();
         }
+
+        /// <summary>
+        /// Returns the prop if locally defined, or null otherwise.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        public ElementSchemaProp GetOwn(string name)
```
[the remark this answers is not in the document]
jsundra: Yeah I wasn't the biggest fan of it. The issue I was running into was not if schema existed or not, but what type an entry could be. How about something like `public Type GetOwnType(string name)`? And then that would either return a type or null, so ElementSchemaProp keeps its guarantees?

Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
jsundra: I added it just to have parity with subscriptions. Right now `ContentWidgetJs` will generate a new JsApi for Audio/Material/Animator for its underlying Asset, if it has the functionality. So there definitely isn't the explicit guarantee of when this would get around to being called.

In play mode, this doesn't really matter, but in edit mode with an Asset able to refresh and update, the subscriptions would end up lingering and leaking. I can update it to use a `Setup`/`Teardown` flow though so it is a lot more explicit with what is happening when.

Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
jsundra: I'll add it in, but I'll keep the logger warnings still so it'd appear in the console for someone potentially using it in scripting.

Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
jsundra: Yeah definitely! I was mulling that over a bit too. That, or maybe a better way to potentially tie some of our APIs together. Something like `tween.material(this, '_Alpha').<...>`. While that doesn't make it quite first class to Schema, it at least makes it more hidden from the scripter until they need to do even more advanced things.

Assets/Source/Player/Scripting/Interface/IAnimator.cs
jsundra: Yeah, I wasn't too happy about creating so much boilerplate without really putting time into figuring out if we'd want other interfaces/conventions, etc. But it did make testing easier, Material was gross in that I was originally having a `new Material(Shader.Find("Unlit/Texture"))` just to get a "blank" Material. Audio throws a ton of errors while trying to set Volume with no source clip too. So yeah, good to have for testing, but thrown in quickly :/

Assets/Source/Player/IUX/Element/ElementSchema.cs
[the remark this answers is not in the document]
thegoldenmule: Dang. Hmmm... I think I like what you have here better than `GetOwnType`. What about `IList<ElementSchemaProp> GetOwnProps()`? This would better mirror dynamic languages.

Assets/Source/Player/Scripting/Interface/Elements/AnimatorJsApi.cs
thegoldenmule: I guess my point was, I'm not sure this can ever be called without unsubscribing those events, as they will keep the object alive.

Assets/Source/Player/IUX/Element/ElementSchema.cs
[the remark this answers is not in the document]
jsundra: Updated to provide a `GetOwnProps` method. I made the return type `ReadOnlyCollection<ElementSchemaProp>`, since it seemed safer/potentially bug-avoiding down the road.

Assets/Source/Player/IUX/Element/ElementSchema.cs
```diff
@@ -462,6 +463,16 @@ public bool HasOwnProp(string name)
 
             return false;
         }
+        
+        /// <summary>
+        /// Returns a ReadOnlyCollection of all props specific to this Element.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        public ReadOnlyCollection<ElementSchemaProp> GetOwnProps()
```
thegoldenmule: Yeeeeees, I think this is much nicer. Keeps all the guarantees and gives you what you want.

Assets/Source/Player/Scripting/Interface/Elements/AudioJsApi.cs
```diff
@@ -29,27 +35,47 @@ public AudioJsApi(ElementSchema schema, IAudioSource audio)
...
+
+            _setup = true;
         }
 
         /// <summary>
-        /// Destructor.
+        /// Unsubscribes from schema.
         /// </summary>
-        ~AudioJsApi()
+        public void Teardown()
```
thegoldenmule: This lifecycle is great. Nice to have guarantees.

PR #224, opened 2018-12-11
Feature/ek 1107 ping aws

jsundra:
- Added a new `NetworkConnectivity` class to measure ping and overall internet access
- Added a new `NetworkJsApi` to expose ping & internet access to the scripting interface.
- Added PingConfig to NetworkConfig.
- Changed SystemJsApi dependencies to be passed through a single `SetDependencies` call.

Right now, pinging can be started with some override config:
```
"Ping": {
    "Enabled": true,
    "Interval": 5
}
```
...or via scripting:
```
system.network.pingEnabled = true;
```
The interval & region can also be configured via scripting. Scripting also has access to a bool for `system.network.online`. 

The method of calculating ping isn't the best, since it includes not just raw RTT time but some overhead of callbacks, etc. Should be good enough for now though.

### DIFF
--- a/Assets/Source/Application/ApplicationConfig.cs
+++ b/Assets/Source/Application/ApplicationConfig.cs
@@ -503,7 +503,7 @@ namespace CreateAR.EnkluPlayer
         /// <summary>
         /// The PingConfig to use.
         /// </summary>
-        public readonly PingConfig Ping = new PingConfig();
+        public PingConfig Ping = new PingConfig();
 
         /// <summary>
         /// Current environment we should connect to.
@@ -756,7 +756,7 @@ namespace CreateAR.EnkluPlayer
         /// <summary>
         /// Whether pings should be sent or not.
         /// </summary>
-        public bool Enabled = false;
+        public bool Enabled = true;
         
         /// <summary>
         /// The interval to send pings, measured in ms.
--- a/Assets/Source/Application/ApplicationConfig.cs
+++ b/Assets/Source/Application/ApplicationConfig.cs
@@ -501,6 +501,11 @@ namespace CreateAR.EnkluPlayer
         public string ApiVersion = "0.0.0";
 
         /// <summary>
+        /// The PingConfig to use.
+        /// </summary>
+        public readonly PingConfig Ping = new PingConfig();
+
+        /// <summary>
         /// Current environment we should connect to.
         /// </summary>
         public string Current;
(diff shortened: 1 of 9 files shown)


Assets/Source/Application/ApplicationConfig.cs
```diff
@@ -500,6 +500,11 @@ public class NetworkConfig
         /// </summary>
         public string ApiVersion = "0.0.0";
 
+        /// <summary>
+        /// The PingConfig to use.
+        /// </summary>
+        public readonly PingConfig Ping = new PingConfig();
```
thegoldenmule: Why readonly? That doesn't seem right.

Assets/Source/Application/Injection/EnkluPlayerModule.cs
```diff
@@ -684,6 +684,13 @@ private void SetupLogging(InjectionBinder binder, LogAppConfig config)
 
                 var appData = binder.GetInstance<IAdminAppDataManager>();
                 binder.Bind<IAppDataManager>().ToValue(appData);
+                
+                SystemJsApi.SetDependencies(
+                    binder.GetInstance<IDeviceMetaProvider>(),
+                    binder.GetInstance<NetworkConnectivity>(),
+                    binder.GetInstance<IMessageRouter>(),
+                    binder.GetInstance<ApiController>(),
+                    binder.GetInstance<ApplicationConfig>());
```
thegoldenmule: Fine with the messiness but let's name it similarly to other stuff like this: `Initialize`.

Assets/Source/Network/NetworkConnectivity.cs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                var startTime = Time.realtimeSinceStartup;
+                var inflight = true;
+                
+                // Send the request
+                _http.Download(url).OnSuccess(httpResponse =>
+                {
+                    Online = true;
+                    PingMs = (Time.realtimeSinceStartup - startTime) * 1000;
+                    _pingMetric.Value(PingMs);
```
thegoldenmule: Simple and effective. I don't think we need anything more exact than this.

Assets/Source/Player/Scripting/Interface/App/NetworkJsApi.cs
```diff
@@ -0,0 +1,67 @@
...
+            set { _networkConnectivity.PingRegion = value; }
+        }
+
+        /// <summary>
+        /// Interval to ping, measured in seconds.
+        /// </summary>
+        public float pingInterval
+        {
+            get { return _networkConnectivity.PingInterval; }
+            // Add a slight minimum so scripting can't spam the network too much.
```
thegoldenmule: :+1:

jsundra:
@thegoldenmule should be good to go!

PR #226, opened 2018-12-12
Feature/metrics

jsundra:
**MetricsUpdateService**
Created a service to track specific metric datapoints that are useful for performance / experience analysis. Currently it is responsible for pushing battery & session duration values.

**PrimaryAnchorManager**
Now sends a metric when individual anchors are specifically unlocated. Importing/Loading/Ready all report 0. 1 is sent when an anchor is unlocated. These metrics could be useful to determine if a specific anchor is consistently unlocated. I added a static function to `MetricsKeys` to handle formatting an anchor's name into a safe metric key. That could be done in the anchor manager itself. Curious what your thoughts are on that one @thegoldenmule.

**PlayApplicationState**
Added a guard to make sure the timer used isn't overwritten. According to the Unity docs, there are multiple states of unlocatedness, so the guard ensures if the device is flip/flopping between different unlocated states that we're not losing tracking time by replacing the timer.

### DIFF
--- a/Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
+++ b/Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
@@ -471,16 +471,6 @@ namespace CreateAR.EnkluPlayer
                 {
                     Messages_OnAutoExport(anchor);
                 }
-
-                // Ensure we're only subbing once
-                Anchor.OnLocated -= Anchor_OnLocatedChanged;
-                Anchor.OnUnlocated -= Anchor_OnLocatedChanged;
-                
-                Anchor.OnLocated += Anchor_OnLocatedChanged;
-                Anchor.OnUnlocated += Anchor_OnLocatedChanged;
-                
-                // Proc the event to push the current state
-                Anchor_OnLocatedChanged(Anchor);
             }
 
             if (null != Anchor)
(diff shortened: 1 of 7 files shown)


Assets/Source/Application/States/Play/Hmd/PrimaryAnchorManager.cs
```diff
@@ -996,5 +1006,24 @@ private void ExportService_OnFileUrlChanged(string url)
...
+            var status = anchor.Status;
+            var name = anchor.Schema.Get<string>("name").Value;
+            switch (status)
+            {
+                case WorldAnchorWidget.WorldAnchorStatus.IsReadyLocated:
+                    _metrics.Value(MetricsKeys.ANCHOR_STATE_UNLOCATED_NAMED(name)).Value(1);
+                    break;
+                case WorldAnchorWidget.WorldAnchorStatus.IsReadyNotLocated:
+                    _metrics.Value(MetricsKeys.ANCHOR_STATE_UNLOCATED_NAMED(name)).Value(0);
+                    break;
```
thegoldenmule: I like the idea here but graphite doesn't really support this method of metrics gathering. Graphite is a static size database with a pretty low upper bound for unique metrics (think like 1000). We only have so many unique metric ids and if you have a bunch of experiences that will use them all up really quickly. I would suggest we table this particular metric and instead log warnings that include the anchor name or id. These will show up in our logs and we can track them down. A long term solution will be having multiple types of analytics gathering. Graphite is just not built for this particular use case.

jsundra: Ah yeah, totally makes sense. I knew it'd be stretching graphite's standard usage, but if we have a cap for unique IDs, we'll hit that way too quickly. I'll remove those bits.

thegoldenmule: :+1: should have communicated this, sorry.

PR #228, opened 2018-12-14
AppDataLoader timeout

jsundra:
Targeting 0.14.0, not sure if you wanted to sneak this in before pushing the build out or not.

NetworkConfig:
- Added "DiskFallbackTime". When nonzero, AppDataLoader will use this delay before ignoring network loading.

AppDataLoader:
- Added a delay that starts when network loading begins. If all of the network data hasn't finished, the loader falls back to loading from disk and ignores any late responses from the original network load attempts.

This isn't perfect - there's a chance that 1/3 or 2/3 of the network requests will finish before the delay trips. The successful network loads shouldn't be used, since they might contain IDs that the remaining data from disk doesn't know about, or vice verse. The disk fallback will load all 3 requests from disk again, and resend the various `RECV_X_LIST` messages, but it'd probably be better if the network requests didn't send a message in that case.

As far as testing goes, the normal loading works still. I got a variety of test scenarios on a saturated network (thank you GTA5 for being a giant blob to download over time). Both cases of no network data coming in, or partial network data coming in first, resulted in successful loads and everything seemed okay.

### DIFF
--- a/Assets/Source/Application/States/Play/App/AppDataLoader.cs
+++ b/Assets/Source/Application/States/Play/App/AppDataLoader.cs
@@ -1,14 +1,17 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using CreateAR.Commons.Unity.Async;
+using CreateAR.Commons.Unity.Http;
 using CreateAR.Commons.Unity.Logging;
 using CreateAR.Commons.Unity.Messaging;
 using CreateAR.EnkluPlayer.IUX;
 using CreateAR.Trellis.Messages;
 using CreateAR.Trellis.Messages.GetPublishedAssets;
+using UnityEngine;
 using Response = CreateAR.Trellis.Messages.GetPublishedScene.Response;
 using Void = CreateAR.Commons.Unity.Async.Void;
 
@@ -17,6 +20,27 @@ namespace CreateAR.EnkluPlayer
     /// <inheritdoc />
     public class AppDataLoader : IAppDataLoader
     {
+        /// <summary>
+        /// The various states the loader will be in.
+        /// </summary>
+        private enum LoadState
+        {
+            /// <summary>
+            /// Nothing, awaiting external load requests.
+            /// </summary>
+            Idle, 
+            
+            /// <summary>
+            /// Loading is via the intended load behavior.
+            /// </summary>
+            Primary, 
+            
+            /// <summary>
+            /// Loading timed out and is loading from disk.
+            /// </summary>
+            TimedOut
+        }
+        
         /// <summary>
         /// Json.
         /// </summary>
@@ -53,9 +77,19 @@ namespace CreateAR.EnkluPlayer
         private readonly HttpRequestCacher _helper;
 
         /// <summary>
-        /// Play mode configuration.
+        /// Network config.
         /// </summary>
-        private PlayAppConfig _config;
+        private readonly NetworkConfig _networkConfig;
+
+        /// <summary>
+        /// IBootstapper.
+        /// </summary>
+        private readonly IBootstrapper _bootstrapper;
+
+        /// <summary>
+        /// The current load state.
+        /// </summary>
+        private LoadState _loadState;
 
         /// <inheritdoc />
         public string Name { get; private set; }
@@ -78,12 +112,16 @@ namespace CreateAR.EnkluPlayer
             ApiController api,
             HttpRequestCacher cache,
             UserPreferenceService prefs,
-            IMessageRouter messages)
+            IMessageRouter messages,
+            IBootstrapper bootstrapper,
+            NetworkConfig networkConfig)
         {
             _api = api;
             _helper = cache;
             _prefs = prefs;
             _messages = messages;
+            _bootstrapper = bootstrapper;
+            _networkConfig = networkConfig;
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


Assets/Source/Application/States/Play/App/AppDataLoader.cs
```diff
@@ -204,11 +242,68 @@ public void Unload()
...
+                {
+                    rtnToken.Fail(exception);
+                }
+            }).OnFinally(_ =>
+            {
+                _loadState = LoadState.Idle;
+            });
+
+            // Set delay before giving up on the network load
+            if (_networkConfig.DiskFallbackTime > 0 && behavior == HttpRequestCacher.LoadBehavior.NetworkFirst)
```
thegoldenmule: Compare against epsilon.

Assets/Source/Application/States/Play/App/AppDataLoader.cs
```diff
@@ -204,11 +242,68 @@ public void Unload()
...
+            yield return new WaitForSeconds(seconds);
+
+            if (_loadState == LoadState.Idle)
+            {
+                // Primary download finished in time!
+                yield break;
+            }
+            
+            Log.Warning(this, "Timeout waiting for network content. Loading from disk.");
+            _loadState = LoadState.TimedOut;
```
thegoldenmule: Part of the benefit of using tokens is that you don't have to do this kind of thing. This probably works, but it's a tad hard to follow. What I gather is that you are setting `_loadState` here and then checking it in the OnSuccess and OnFailure callbacks. Instead, you could just keep a reference to the token returned by `Async.All` and call `Abort` here. Then OnSuccess and OnFailure wouldn't even be called. It's stylistic, again I think either works, but it is part of the point of tokens to simplify such a thing.

Assets/Source/Application/States/Play/App/AppDataLoader.cs
```diff
@@ -204,11 +242,68 @@ public void Unload()
...
+            Async.All(
+                LoadPrerequisites(appId, HttpRequestCacher.LoadBehavior.DiskOnly),
+                LoadScenes(appId, HttpRequestCacher.LoadBehavior.DiskOnly)).OnSuccess(_ =>
+            {
+                rtnToken.Succeed(Void.Instance);
+            }).OnFailure(rtnToken.Fail)
+              .OnFinally(_ =>
+            {
+                _loadState = LoadState.Idle;
+            });
```
thegoldenmule: Again, line breaks before the `.` I think provides better readability and cohesion with the rest of the codebase.

thegoldenmule:
@jsundra Another thing to consider: what about moving the timeout into the `HttpRequestCacher` and each request could have a shorter timeout? Then `AppDataLoader` wouldn't have to change much.

Assets/Source/Application/States/Play/App/AppDataLoader.cs
```diff
@@ -204,11 +242,68 @@ public void Unload()
...
+            }).OnFailure(exception =>
+            {
+                if (_loadState == LoadState.Primary)
+                {
+                    rtnToken.Fail(exception);
+                }
+            }).OnFinally(_ =>
+            {
+                _loadState = LoadState.Idle;
+            });
```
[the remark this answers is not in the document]
jsundra: I actually like that better too! Not sure why I didn't do it the first time around

Assets/Source/Application/States/Play/App/AppDataLoader.cs
jsundra: Ah, yeah that'd be a really nice way to clean up the network loads without having to check state in their success. I tried looking into this, but I don't think it's possible to remove `_loadState` fully currently. The biggest issue being that I don't think there's a cleaner way to have the delay coroutine know that it shouldn't trigger the fallback load.

Similarly, cleaning up the LoadApp function would be possible if there was something like `Async.Any` or `Async.Race`. I think the cleanest way it could look is if there was some exclusive race that took a number of promises and when one resolved, aborted the others.

I'll sleep on it and see if I can find a way to clean things up in the morning.

jsundra:
Hmm. It'd be great to have the logic out of `AppDataLoader`, especially since the http handling has behaviors built directly into it. I think I'd be cautious adding it directly into `HttpRequestCache` right now though. There might be false alarms where a certain request just takes a long time by nature, triggers the time threshold, then forces the rest of loading to be local-only. It'd also be difficult at that layer for cases like `AppDataLoader` that need 3 requests. If the last falls back to disk but returns successful, 2/3 could come from the server and mismatch the 3rd request.

Assets/Source/Application/States/Play/App/AppDataLoader.cs
thegoldenmule: How about this:

```
_networkToken = Async.All(...);
_networkToken.OnFinally(_ => _networkToken = null);


// coroutine
yield return new WaitForSecondsRealtime(TIMEOUT);

if (null == _networkToken) {
	return;
}

_networkToken.Abort();
_networkToken = null;
```

jsundra: I think that works, but you'd have to cast the `IAsyncToken` coming out of `Async.All` to `AsyncToken`. Do-able, just cringe worthy.

jsundra: Giving this a try - Unless I'm missing something I don't think aborting the token returned from `Async.All` works. I tried with a simplified setup:
```
var network1 = new AsyncToken<Void>();
var network2 = new AsyncToken<Void>();

_bootstrapper.BootstrapCoroutine(Delay(1, network1));
_bootstrapper.BootstrapCoroutine(Delay(2, network2));

var allToken = (AsyncToken<Void[]>) Async.All(network1, network2);
allToken.OnSuccess(_ => Log.Warning(this, "All token finished"));

allToken.Abort();
```
```
private IEnumerator Delay(float seconds, AsyncToken<Void> callback)
{
    callback.OnSuccess(_ => Log.Warning(this, "Delay Proc!"));
    
    yield return new WaitForSeconds(seconds);
    
    callback.Succeed(Void.Instance);
}
```

When running this, `Abort()` stops `allToken` from succeeding/logging, but the underlying network tokens will still succeed and log.

jsundra:
@thegoldenmule ready for another round!

In the future we could remove a lot of the token wrapping if we add something like `Async.Race` and a util for delaying a Token. `LoadApp` would then look cleaner, something like this:
```
return Async.Race(
  // Primary load
  Async.All(
    LoadPrerequisites(appId, behavior),
    LoadScenes(appId, behavior)),
  // Timeout load
  Async.Delay(_networkConfig.DiskFallbackSecs, Async.All(
    LoadPrerequisites(appId, HttpRequestCacher.LoadBehavior.DiskOnly),
    LoadScenes(appId, HttpRequestCacher.LoadBehavior.DiskOnly))),
).OnFinally(_ => InvalidateTokens());
```

Assets/Source/Application/States/Play/App/AppDataLoader.cs
```diff
@@ -328,18 +297,14 @@ private IAsyncToken<Void> LoadPrerequisites(string appId, HttpRequestCacher.Load
...
                     "appdata://" + appId + "/scenelist",
                     () => _api.PublishedApps.GetPublishedApp(appId))
                 .OnSuccess(response =>
                 {
-                    if (behavior == HttpRequestCacher.LoadBehavior.NetworkFirst && _loadState != LoadState.Primary)
-                    {
-                        Log.Warning(this, "Late network response. Ignoring.");
-                        return;
-                    }
+                    Log.Warning(this, "response!");
```
thegoldenmule: Looks like you forgot to take out a log?

thegoldenmule:
Looks good to me. Those util methods would be interesting. I'm game (for later) as long as we have strong guarantees on semantics.

PR #229, opened 2018-12-18
Proximity Spam Reduction

jsundra:
Remove some needless spam from the proximity system


jsundra:
Honestly I think I'd still rather remove it than switch it to debug. If there's only 1-2 elements in an experience it's okay, but once a handful of things use the proximity system, it really just becomes useless :/

PR #230, opened 2018-12-20
EK-1159: Don't notify client of events that originated in the client.

Pixelwelder:
[EK-1159](https://enkluplay.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=EK&modal=detail&selectedIssue=EK-1159)

`RTObjectSelection.ClearSelection` and `RTObjectSelection.SetSelectedObjects` are both only used by the `IDesignController`. If they are intended for more general use, we can pass in a reason instead of setting one inside these two.

Another possible improvement would be using a source AND a reason.

If I misinterpreted intent, please let me know.


thegoldenmule:
Looks good to me. Have you tested that selection and deselection work from both sides of the bridge?

Pixelwelder:
> Looks good to me. Have you tested that selection and deselection work from both sides of the bridge?

I have indeed.

PR #232, opened 2019-01-04
EK-792: Transfer initial render settings from editor

Pixelwelder:


Assets/Source/Application/Bridge/EditorSettings.cs
```diff
@@ -0,0 +1,230 @@
...
+using CreateAR.Commons.Unity.Logging;
+using CreateAR.EnkluPlayer.IUX;
+using RLD;
+using Source.Messages.ToApplication;
+using Object = UnityEngine.Object;
+
+namespace CreateAR.EnkluPlayer
+{
+    public class EditorSettings
+    {
```
thegoldenmule: We should have a comment on every class.

Assets/Source/Application/Bridge/EditorSettings.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// <summary>
+        /// Fired when the value of any setting has changed.
+        /// </summary>
+        public event Action<SettingChangedArgs> OnChanged;
+
+        /// <summary>
+        /// All scenes.
+        /// </summary>
+        [Inject]
+        public IAppSceneManager Scenes { get; set; }
```
thegoldenmule: The inject attribute is really only for monobehaviours. This should just be passed in through the constructor.

Assets/Source/Application/Bridge/EditorSettings.cs
```diff
@@ -0,0 +1,230 @@
...
+                
+                var scans = new List<Element>();
+                var all = Scenes.All;
+                
+                foreach (var id in all)
+                {
+                    var root = Scenes.Root(id);
+                    root.Find("..(@type==ScanWidget)", scans);
+                    
+                    foreach (var scan in scans)
```
thegoldenmule: Likely this foreach allocates. All foreaches should be switched to fors. More info: https://stackoverflow.com/questions/18552669/memory-allocation-when-using-foreach-loops-in-c-sharp

Assets/Source/Application/Bridge/EditorSettings.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        /// <summary>
+        /// Sets any setting.
+        /// </summary>
+        public void SetSetting(string name, bool value)
+        {
+            Log.Info(this, "Setting {0} to {1}", name, value);
+            switch (name)
+            {
+                case "MeshScan":
```
thegoldenmule: In our c# standard braces are not optional.

Assets/Source/Application/Bridge/EditorSettings.cs
```diff
@@ -0,0 +1,230 @@
...
+        Grid,
+        ElementGizmos,
+        HierarchyLines,
+        All
+    }
+    
+    /// <summary>
+    /// The args for a settings change event.
+    /// </summary>
+    public class SettingChangedArgs : EventArgs
```
thegoldenmule: We haven't used the EventArgs pattern elsewhere so I'd like to keep it out of our classes.

Assets/Source/Application/Bridge/EditorSettings.cs
thegoldenmule: (Constructor injection is safer)

Assets/Source/Application/Services/ElementActionHelperService.cs
```diff
@@ -166,74 +173,7 @@ private void OnSetting(BridgeHelperSettingsEvent @event)
...
-                    }
-                    default:
-                    {
-                        Log.Warning(this,
-                            "Could not set render property {0}. No such property exists.",
-                            @event.Name);
-                        break;
-                    }
-                }
+                _editorSettings.SetSetting(@event.Name, @event.Value);
```
thegoldenmule: Consider renaming to `Set`, as you already know the class deals with settings exclusively.

Assets/Source/Application/States/Play/Desktop/DesktopDesignController.cs
```diff
@@ -419,6 +411,17 @@ private void Editor_OnSelectionChanged(ObjectSelectionChangedEventArgs args)
             }
         }
 
+        private void Editor_OnSettingsChanged(SettingChangedArgs args)
```
thegoldenmule: Missing comment.

Assets/Source/Application/States/Play/Hmd/Controllers/HmdDesignController.cs
```diff
@@ -80,10 +81,10 @@ public class HmdDesignController : IDesignController
         private GameObject _referenceCube;
 
         /// <summary>
-        /// The mode of this design controller.
+        /// Contains editor settings.
         /// </summary>
-        private DesignControllerMode _mode = DesignControllerMode.Normal;
-        
+        private EditorSettings _editorSettings;
```
thegoldenmule: Can probably be readonly

Assets/Source/Application/States/Play/Hmd/Controllers/HmdDesignController.cs
```diff
@@ -555,5 +551,13 @@ private void Voice_OnPlay(string command)
 
             _messages.Publish(MessageTypes.LOAD_APP);
         }
+
+        private void Editor_OnSettingsChanged(SettingChangedArgs args)
```
thegoldenmule: comment

Assets/Source/Application/States/ReceiveApp/ReceiveAppApplicationState.cs
```diff
@@ -2,6 +2,8 @@
 using CreateAR.Commons.Unity.Http;
 using CreateAR.Commons.Unity.Logging;
 using CreateAR.Commons.Unity.Messaging;
+using Source.Messages.ToApplication;
```
thegoldenmule: One of your new classes has a non-conforming namespace. This should be renamed to match the codebase.

Assets/Source/Application/States/ReceiveApp/ReceiveAppApplicationState.cs
```diff
@@ -180,6 +191,27 @@ private void WaitForAppInfo(Action callback)
...
+            Log.Info(this, "Waiting on editor settings.");
+
+            _messages.SubscribeOnce(
+                MessageTypes.INITIAL_EDITOR_SETTINGS,
+                obj =>
+                {
+                    var settingsEvent = (EditorSettingsEvent) obj;
+                    Log.Info(this, "Editor settings received : {0}.", settingsEvent);
+
+                    _editorSettings.PopulateFromEvent(settingsEvent);
```
thegoldenmule: Consider the more concise name `Populate`. Adding the type to the function name conveys no relevant or additional info-- i.e. you already know it takes an event.

Assets/Source/Messages/ToApplication/EditorSettingsEvent.cs
```diff
@@ -0,0 +1,45 @@
+using Newtonsoft.Json;
+
+namespace Source.Messages.ToApplication
```
thegoldenmule: Here it is. Rider can rename this for you.

Assets/Source/Messages/ToApplication/EditorSettingsEvent.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    /// <summary>
+    /// Event containing initial settings from the editor.
+    /// </summary>
+    public class EditorSettingsEvent
+    {
+        /// <summary>
+        /// Whether the mesh scan is visible.
+        /// </summary>
+        [JsonProperty("MeshScan")]
```
thegoldenmule: JsonProperty is only necessary when the name is different than the field name.

Assets/Source/Application/Bridge/EditorSettings.cs
Pixelwelder: Any idea why Unity crashes when I inject via constructor?

thegoldenmule: Usually Strange catches loops but this might be caused by a circular dependency.

Assets/Editor/Test/IUX/DummyElementFactory.cs
```diff
@@ -14,6 +14,12 @@ public Element Element(string vine)
         {
             throw new NotImplementedException();
         }
+
+        /// <inheritdoc />
+        public void SetOverride(int type, ElementSchema schema)
+        {
+            throw new NotImplementedException();
```
Pixelwelder: ¯\_(ツ)_/¯

Pixelwelder: ¯\\_(ツ)_/¯

Assets/Source/Application/States/Play/Desktop/DesktopDesignController.cs
```diff
@@ -33,6 +33,11 @@ public class DesktopDesignController : IDesignController
         /// Primary anchor.
         /// </summary>
         private readonly IPrimaryAnchorManager _primaryAnchor;
+        
+        /// <summary>
+        /// Contains editor settings.
+        /// </summary>
+        private readonly EditorSettings _editorSettings;
```
Pixelwelder: If you feel strongly about it, I could remove this and inject a value externally instead of using an event-based approach.

Assets/Source/Application/States/Play/Hmd/Controllers/HmdDesignController.cs
```diff
@@ -120,6 +126,7 @@ public IAppSceneManager Scenes
             IMessageRouter messages,
             IPrimaryAnchorManager primaryAnchor,
             ApiController api,
+            EditorSettings editorSettings,
```
Pixelwelder: This could also be removed if we we go for direct calls over events.

Assets/Source/Application/Bridge/EditorSettings.cs
```diff
@@ -0,0 +1,251 @@
...
+
+        /// <summary>
+        /// Requests the utility to run through all values and update the necessary components.
+        /// </summary>
+        public void Update()
+        {
+            MeshScan = MeshScan;
+            Grid = Grid;
+            ElementGizmos = ElementGizmos;
+            HierarchyLines = HierarchyLines;
```
thegoldenmule: I must have missed this part before-- but I think this is a dangerous pattern. It would be preferable to pull the application logic out of the properties (like `MeshScan`) and into methods (like `UpdateMeshScan`). Then this method and the properties would call those methods.

Assets/Source/Player/IUX/Element/IElementFactory.cs
```diff
@@ -13,10 +13,17 @@ public interface IElementFactory
...
         /// <param name="vine">The raw vine.</param>
         /// <returns></returns>
         Element Element(string vine);
+
+        /// <summary>
+        /// Sets a schema override. This schema will be added on top of all other schemas for the specified type.
+        /// </summary>
+        /// <param name="type">The ElementType that will receive the schema.</param>
+        /// <param name="schema">The ElementSchema that will be given to the elements of the specified type.</param>
+        void SetOverride(int type, ElementSchema schema);
```
thegoldenmule: There is some existing tech debt you're running into here-- `ElementFactory` should not be setting up element base schema to begin with. This pushes it over the edge for me into refactor time. I think there should be a new object that sets up the base schema (just extract those portions of the class). That object is then injected into both the factory and the settings object. That new object then has a nice interface for mucking with base schema.

Assets/Source/Application/States/Play/Desktop/DesktopDesignController.cs
thegoldenmule: Either is acceptable for me.

PR #235, opened 2019-01-14
EK-792: Transfer initial settings from editor to player.

Pixelwelder:
This is a new PR since it branches off the earlier approach.


Assets/Source/Application/Bridge/EditorSettings.cs
```diff
@@ -0,0 +1,226 @@
...
+
+        /// <summary>
+        /// Requests the utility to run through all values and update the necessary components.
+        /// </summary>
+        public void Update()
+        {
+            MeshScan = MeshScan;
+            Grid = Grid;
+            ElementGizmos = ElementGizmos;
+            HierarchyLines = HierarchyLines;
```
thegoldenmule: Same comments here as last time-- this pattern is dangerous.

PR #237, opened 2019-01-16
Feature/video capture

jsundra:
Since VideoMode and PhotoMode are exclusive, I moved a lot of the functionality from `SnapshotCapture` -> `MediaCapture`, and added the support for video recording. I removed the custom width/height setting for images, since there's a very explicit table of what values are supported, especially with video in the mix. It can be revisited later if we need low res/framerate support.

MediaCapture supports handling interactions thru AsyncTokens, so all the async nature of the underlying UWP.

Also included is a jsapi for interacting with `IMediaCapture`. Right now it's available under `system.device.media.*`. Not a huge fan of how lengthy that is, but it also seems like a good spot for it to live.


Assets/Source/Application/States/Play/IMediaCapture.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// Exits VideoMode.
+        /// </summary>
+        IAsyncToken<Void> ExitVideoMode();
+
+        /// <summary>
+        /// Shuts down the capture.
+        /// </summary>
+        void Teardown();
+    }
+}
```
thegoldenmule: This is a long interface, which means it might be too complicated. A few things:

* I am doubtful that this will translate well across platforms. iOS and Android do not work this way, for instance.
* It requires users to know a _lot_ about the implementation details, and it's easy to use wrong (like taking a pic while video mode is on).
* It requires users to know about, say, existing MS bugs. 

I would recommend getting rid of the enter/exit modes and pulling them inside the other methods, i.e. taking a picture or video automatically starts/stops the appropriate mode.

Assets/Source/Application/States/Play/Hmd/MediaCapture.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// </summary>
+        private VideoCapture _videoCapture;
+        
+        /// <inheritdoc />
+        public CaptureState CaptureState { get; private set; }
+
+        /// <inheritdoc />
+        public void Setup()
+        {
+            // TODO: if Teardown() is called before this returns, we need to destroy the
```
thegoldenmule: You already have a `CaptureState`, use it here to check if we've already been setup or if we need to cleanup.

Assets/Source/Application/States/Play/IMediaCapture.cs
thegoldenmule: Another note: it's odd that I have to call `Setup` and `Teardown` but then also call another setup/teardown function (the mode methods). You should also pull these inside the other methods-- I see no need to have the user have to worry about them.

Assets/Source/Application/States/Play/Hmd/MediaCapture.cs
```diff
@@ -0,0 +1,310 @@
...
+            var cameraParameters = new CameraParameters
+            {
+                hologramOpacity = 1.0f,
+                cameraResolutionWidth = WIDTH,
+                cameraResolutionHeight = HEIGHT
+            };
+
+            var rtnToken = new AsyncToken<Void>();
+            
+            _photoCapture.StartPhotoModeAsync(cameraParameters, result =>
```
thegoldenmule: Are you sure these async methods return on the main thread?

Assets/Source/Application/States/Play/Hmd/MediaCapture.cs
```diff
@@ -0,0 +1,310 @@
...
+                CaptureState = CaptureState.VideoMode;
+                
+                rtnToken.Succeed(Void.Instance);
+            });
+            
+            return rtnToken;
+        }
+        
+        /// <inheritdoc />
+        public IAsyncToken<Void> StartRecording()
```
thegoldenmule: It seems odd that I can't specify the path to record to, or get one back.

Assets/Source/Application/States/Play/PlayApplicationState.cs
```diff
@@ -170,6 +173,9 @@ public void Enter(object context)
             _voice.Register("reset", Voice_OnReset);
             _voice.Register("update", Voice_OnUpdate);
             
+            // initialize media capture
+            _mediaCapture.Setup();
```
thegoldenmule: An additional thing we should keep in mind: we should not force the application to consume resources it doesn't need. Here, for instance, we're creating the capture objects even if you never use it. That's another reason to pull setup and teardown inside the implementation.

Assets/Source/Player/Scripting/Interface/App/MediaCaptureJsApi.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// <summary>
+        /// Temp for debugging.
+        /// </summary>
+        public void rebuild()
+        {
+            _mediaCapture.Teardown();
+            _mediaCapture.Setup();
+        }
+    }
+}
```
thegoldenmule: The js interface could use a similar cleanup.

Assets/Source/Application/Injection/EnkluPlayerModule.cs
```diff
@@ -528,11 +528,7 @@ private void SetupLogging(InjectionBinder binder, LogAppConfig config)
 
             // Camera
             {
-#if NETFX_CORE
-                binder.Bind<ISnapshotCapture>().To<SnapshotCapture>().ToSingleton();
-#else
-                binder.Bind<ISnapshotCapture>().To<PassthroughSnapshotCapture>().ToSingleton();
-#endif
+                binder.Bind<IMediaCapture>().To<MediaCapture>().ToSingleton();
```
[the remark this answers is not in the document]
jsundra: The `VideoCapture` API in other build targets like WebGL just returns unsuccessful attempts, but yeah it'd be cleaner now to have a passthrough for non-hololens targets. Will add!

Assets/Source/Application/States/Play/IMediaCapture.cs
jsundra: Definitely agree. I originally wrote the interface with just a Start/Stop for video, and a single function for capturing images. It works and is pretty concise, and is better suited for cross platform implementations. The downside to that though for HoloLens is that the amount of time to create the `VideoCapture` instance, enter VideoMode, and then start recording starts to stack up.

Maybe the solution is to have all the steps run if needed when calling `StartRecording`, but to still provide a separate call for something like `PrepRecording` that handles getting a `VideoCapture` instance and entering VideoMode ahead of time? Still not the most cross-platform friendly, but at least is a lot less boilerplate, and avoids having to know specific implementation details if you just want to use Start/Stop without worrying about latency.

Assets/Source/Application/States/Play/Hmd/MediaCapture.cs
jsundra: Both the Unity docs and the UWP references don't mention anything about threading. I'll try and dig a little deeper and see if there's any more information.

Assets/Source/Application/States/Play/Hmd/HoloLensImageCapture.cs
```diff
@@ -0,0 +1,146 @@
...
+                    else
+                    {
+                        Log.Info(this, "Entered PhotoMode.");
+                        rtnToken.Succeed(Void.Instance);
+                    }
+                });
+            });
+
+            return rtnToken;
+        }
```
thegoldenmule: To be a bit nitpicky-- this implementation has a race condition. I could call `Capture(); Capture(); or `Warm(); Warm();` and it would be calling `PhotoCapture.CreateAsync` twice before the async callback is called. Additionally, `Capture` is checking only that `_photoCapture` exists before use, but that `Capture` could be called before `StartPhotoModeAsync` returns. In general it's inconsistent to count on state set in the callback like this-- especially when there are two callbacks.

Instead, `Warm` could immediately create a token at the class level (like `_warmToken = new AsyncToken()`). Subsequent calls can then just return `_warmToken.Token()`. Then `Capture` would just call `Warm()` instead of having it worry about the internal logic of `Warm`.

Assets/Source/Application/States/Play/Hmd/HoloLensVideoCapture.cs
```diff
@@ -0,0 +1,188 @@
...
+                        else
+                        {
+                            Log.Info(this, "Entered VideoMode.");
+                            rtnToken.Succeed(Void.Instance);
+                        }
+                    });
+            });
+
+            return rtnToken;
+        }
```
thegoldenmule: Same issue here as in photocap. Should use a token.

Assets/Source/Application/States/Play/Hmd/HoloLensVideoCapture.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        /// <inheritdoc />
+        public IAsyncToken<Void> Start()
```
thegoldenmule: I'm still not seeing any way to set or retrieve the path in either implementation.

Assets/Source/Application/States/Play/Hmd/HoloLensVideoCapture.cs
```diff
@@ -0,0 +1,188 @@
...
+            
+            return rtnToken;
+        }
+
+        /// <inheritdoc />
+        public IAsyncToken<Void> Abort()
+        {
+            var rtnToken = new AsyncToken<Void>();
+            
+            if (_videoCapture != null)
```
thegoldenmule: Same issue here, `_videoCapture` is set to null inside an async callback. The simplest solution here is probably just a temp variable:

```
var temp = _videoCapture;
_videoCapture = null;
temp.StopVideoModeAsync(...);
```

Assets/Source/Application/States/Play/IImageCapture.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <returns></returns>
+        IAsyncToken<string> Capture();
+
+        /// <summary>
+        /// Aborts the image capture and frees underlying resources.
+        /// </summary>
+        /// <returns></returns>
+        IAsyncToken<Void> Abort();
+    }
+}
```
thegoldenmule: Much cleaner!

Assets/Source/Application/States/Play/IVideoCapture.cs
```diff
@@ -0,0 +1,40 @@
...
+        IAsyncToken<string> Stop();
+
+        /// <summary>
+        /// Aborts the video capture and frees underlying resources.
+        /// </summary>
+        /// <returns></returns>
+        IAsyncToken<Void> Abort();
+        
+    }
+}
```
thegoldenmule: Mmmmmm simplicity.... :)

Assets/Source/Application/States/Play/Hmd/HoloLensImageCapture.cs
jsundra: Yeah definitely, I'll make this more robust!

jsundra:
@thegoldenmule I think everything is ready for another round of review!

One thing I had to add, which isn't new but still cringeworthy is `startCustomPath` and `captureCustomPath` to get around the jint method overloading bug. Not in love with those names at all, so I'll happily change them if you have a better suggestion :)

PR #240, opened 2019-01-22
Added checksum and hardware id js apis

thegoldenmule:
```javascript
var id = require('checksum').crc16(app.system.device.hardwareId);
```


thegoldenmule:
Docs: https://github.com/enklu/enkluplayer-jsapi-docs/pull/7

PR #241, opened 2019-01-23
Feature/video upload

jsundra:
Exposes creation actions for image/video capture, and introduces `IVideoManager` and `IImageManager` with a HoloLens implementation for each. They ended up not diverging as much as I thought and could probably be combined in the future potentially. But for now I think it's safer to keep them separate in case there's divergent functionality.

There's a new `SnapConfig` now that determines where files should be stored, how many failures should be kept on disk, and how long to wait after failures before attempting again.

The uploaders will always attempt to grab from new images/videos before grabbing previously failed ones.

Added some basic metrics for image/video capture attempts, success, and failure. For uploading, I figure the files will exist or not, and we'll have loggly logs to look into specific upload errors should they arise.


Assets/Source/Application/ApplicationConfig.cs
```diff
@@ -801,4 +806,35 @@ public void Override(DebugConfig config)
...
+        /// <summary>
+        /// The root folder videos will be stored at under the app's datapath.
+        /// </summary>
+        public string VideoFolder = "videos";
+
+        /// <summary>
+        /// The root folder images will be stored at under the app's datapath.
+        /// </summary>
+        public string ImageFolder = "images";
+    }
```
thegoldenmule: Needs an `Override(other)` defined.

Assets/Source/Application/States/Play/Hmd/HoloLensVideoManager.cs
```diff
@@ -0,0 +1,347 @@
...
+                fileStream.Dispose();
+                
+                if (response.Success)
+                {
+                    token.Succeed(Void.Instance);
+                }
+                else
+                {
+                    token.Fail(new Exception(response.Error));
+                }
```
thegoldenmule: shoot me

Assets/Source/Application/States/Play/Hmd/HoloLensVideoManager.cs
```diff
@@ -0,0 +1,347 @@
...
+using Newtonsoft.Json;
+using Void = CreateAR.Commons.Unity.Async.Void;
+
+namespace CreateAR.EnkluPlayer
+{
+    /// <summary>
+    /// VideoManager for HoloLens. Handles uploading videos to Trellis
+    /// and managing failed uploads with a set cap.
+    /// </summary>
+    public class HoloLensVideoManager : IVideoManager
```
thegoldenmule: Does this need to be `HoloLens` specific? Seems like it could just be `VideoManager`.

Assets/Source/Application/States/Play/HoloLensImageManager.cs
```diff
@@ -0,0 +1,347 @@
...
+using Newtonsoft.Json;
+using Void = CreateAR.Commons.Unity.Async.Void;
+
+namespace CreateAR.EnkluPlayer
+{
+    /// <summary>
+    /// ImageManager for HoloLens. Handles uploading images to Trellis
+    /// and managing failed uploads with a set cap.
+    /// </summary>
+    public class HoloLensImageManager : IImageManager
```
thegoldenmule: Same here, this doesn't seem HoloLens specific. `IImageCapture` is your abstraction, all hololens stuff is already squared away inside.

Assets/Source/Application/States/Play/HoloLensImageManager.cs
```diff
@@ -0,0 +1,347 @@
...
+                        }
+                    });
+            }
+
+            return rtnToken;
+        }
+    }
+}
+
+#endif
```
thegoldenmule: Yah let's make a ticket in the future about combining these two. They are basically the same thing.

Assets/Source/Application/States/Play/PassthroughMediaManager.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// Passthrough implementation of IVideoManager and IImageManager.
+    /// </summary>
+    public class PassthroughMediaManager : IVideoManager, IImageManager
+    {
+        /// <summary>
+        /// Enabled uploads.
+        /// </summary>
+        public void EnableUploads(string tag, bool uploadExisting = false)
+        {
+            throw new System.NotImplementedException();
```
thegoldenmule: A passthrough doesn't normally throw-- it does nothing. This would probably cause issues on the web.

Assets/Source/Application/States/Play/Hmd/HoloLensVideoManager.cs
jsundra: Hmm... I suppose in the future no. Back when I was using Windows APIs it definitely would have been. Now the only issue is that I'm using Tasks and HttpClient which aren't .NET 3.5 friendly :/

Assets/Source/Application/States/Play/HoloLensImageManager.cs
jsundra: EK-1322

PR #247, opened 2019-01-25
Refactor HoloLensVideoCapture to be more sane

jsundra:
Switching to use more managed state, and not requiring as explicit a warm/abort process as before.

### DIFF
--- a/Assets/Source/Application/States/Play/Hmd/HoloLensVideoCapture.cs
+++ b/Assets/Source/Application/States/Play/Hmd/HoloLensVideoCapture.cs
@@ -161,9 +161,10 @@ namespace CreateAR.EnkluPlayer
                         }
                         else
                         {
+                            var exception = new Exception(string.Format("Failure entering VideoMode ({0})", result.hResult));
+                            Log.Error(this, exception);
                             _currentState = CaptureState.Error;
-                            rtnToken.Fail(new Exception(string.Format("Failure entering VideoMode ({0})",
-                                result.hResult)));
+                            rtnToken.Fail(exception);
                         }
                     });
             });
@@ -212,7 +213,9 @@ namespace CreateAR.EnkluPlayer
                 else
                 {
                     _currentState = CaptureState.Error;
-                    rtnToken.Fail(new Exception(string.Format("Failure starting recording ({0})", result.hResult)));
+                    var exception = new Exception(string.Format("Failure starting recording ({0})", result.hResult));
+                    Log.Error(this, exception);
+                    rtnToken.Fail(exception);
                 }
             });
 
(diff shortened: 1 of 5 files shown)


Assets/Source/Application/States/Play/Hmd/HoloLensVideoCapture.cs
```diff
@@ -70,213 +118,228 @@ public bool IsRecording
...
-                            Abort();
+                            Log.Info(this, "Entered VideoMode.");
+                            _currentState = CaptureState.InVideoMode;
+                            rtnToken.Succeed(Void.Instance);
                         }
                         else
                         {
-                            Log.Info(this, "Entered VideoMode.");
-                            _bootstrapper.BootstrapCoroutine(SucceedToken(_warmToken, Void.Instance));
+                            _currentState = CaptureState.Error;
```
thegoldenmule: Let's log this case too.

Assets/Source/Application/States/Play/Hmd/HoloLensVideoCapture.cs
```diff
@@ -70,213 +118,228 @@ public bool IsRecording
...
-                    _bootstrapper.BootstrapCoroutine(FailToken(rtnToken, exception));
-                });
+                    _currentState = CaptureState.Recording;
+                    rtnToken.Succeed(Void.Instance);
+                }
+                else
+                {
+                    _currentState = CaptureState.Error;
+                    rtnToken.Fail(new Exception(string.Format("Failure starting recording ({0})", result.hResult)));
+                }
```
thegoldenmule: How about logs for both cases.

Assets/Source/Application/States/Play/Hmd/HoloLensVideoCapture.cs
```diff
@@ -70,213 +118,228 @@ public bool IsRecording
...
-        /// <returns></returns>
-        private IEnumerator FailToken<T>(AsyncToken<T> token, Exception exception)
-        {
-            yield return null;
-            token.Fail(exception);
-        }
+            var rtnToken = new AsyncToken<Void>();
+            
+            Log.Info(this, "Refreshing VideoMode.");
+            _videoCapture.StopVideoModeAsync(stopResult =>
```
thegoldenmule: No null check here scares me.

Assets/Source/Application/States/Play/Hmd/HoloLensVideoCapture.cs
```diff
@@ -70,213 +118,228 @@ public bool IsRecording
...
+                    _videoCapture.StartVideoModeAsync(_cameraParameters, VideoCapture.AudioState.ApplicationAndMicAudio, startResult =>
+                    {
+                        if (startResult.success)
+                        {
+                            Log.Info(this, "VideoMode Refreshed.");
+                            rtnToken.Succeed(Void.Instance);
+                        }
+                        else
+                        {
+                            rtnToken.Fail(new Exception(string.Format("Failure entering VideoMode ({0})",
```
thegoldenmule: log

jsundra:
A lot of those errors would surface in the scripting APIs handling of them - but yeah it's better to have the capture system itself emitting the logs directly, and not sometime later in callbacks that may be out of order. Updated!

PR #248, opened 2019-01-26
VoiceJs!

jsundra:
It's that API we keep saying we should add!


Assets/Source/Player/Scripting/Interface/VoiceJsInterface.cs
```diff
@@ -0,0 +1,135 @@
...
+        
+        /// <summary>
+        /// The backing IVoiceCommandManager.
+        /// </summary>
+        private IVoiceCommandManager _voice;
+        
+        /// <summary>
+        /// Cache of bound commands/callbacks.
+        /// </summary>
+        private Dictionary<string, JsCallback> _callbacks = new Dictionary<string, JsCallback>();
```
thegoldenmule: `readonly`

PR #252, opened 2019-01-31
Feature/debug uis

jsundra:
- Added script download failure to `ApplicationConfig`
- Added public failure lists to `IScriptLoader` and `IAssetLoader`. Scripts don't work off a queue, so the implementation isn't perfect, there's potential for a race condition currently (experience unloading while script is loading)
- Tweaked some schema related accessors.

Experience HUD:
nothing_works

Network HUD:
image

Anchors HUD:
image


Assets/Source/Application/ApplicationConfig.cs
```diff
@@ -319,6 +319,11 @@ public class NetworkConfig
         /// </summary>
         public float AnchorImportFailChance = 0.0f;
 
+        /// <summary>
+        /// Likelyhood of forcing script download to fail.
+        /// </summary>
```
thegoldenmule: * likelihood

Assets/Source/Application/States/Play/PlayApplicationState.cs
```diff
@@ -452,5 +462,65 @@ private void Voice_OnUpdate(string command)
...
+        
+        /// <summary>
+        /// Called when the experience command is recognized.
+        /// </summary>
+        /// <param name="command">The command.</param>
+        private void Voice_OnExperience(string command)
+        {
+            int id;
+            _ui
+                .Open<ExperienceUIView>(new UIReference
```
thegoldenmule: These should probably be OpenOverlay, no?

Assets/Source/Player/App/IScriptLoader.cs
```diff
@@ -1,17 +1,50 @@
...
         /// </summary>
         /// <param name="script">The associated data.</param>
         /// <returns></returns>
         IAsyncToken<string> Load(ScriptData script);
+
+        /// <summary>
+        /// Resets tracking of load failures and resets the queue length.
+        /// </summary>
+        void ResetLoadTracking();
     }
```
thegoldenmule: I'm on the fence about adding these things to the interface. It muddies it quite a bit, requires changes to unrelated implementations, and I'm not sure it's necessary. I think a debug hud should be free to do debuggish things: like cast IScriptLoader to an implementation that has these extra things. What do you think?

Assets/Source/Player/Asset/IAssetLoader.cs
```diff
@@ -13,6 +31,16 @@ public interface IAssetLoader
...
+        
+        /// <summary>
+        /// The number of asset loads in progress.
+        /// </summary>
+        int QueueLength { get; }
+        
+        /// <summary>
+        /// A collection of load failures this IAssetLoader experienced.
+        /// </summary>
+        List<AssetLoadFailure> LoadFailures { get; }
```
thegoldenmule: Same here, I think it's acceptable for a debug hud to cast to a specific implementation.

Assets/Source/Player/App/IScriptLoader.cs
jsundra: Yeah, I agree that shoving debug support into an interface muddies it up. Casting works, though I dislike that it's a weaker contract that will only potentially fail when a debug window opens. Maybe as we add more debug tooling we should invert the relationship and have runtime implementations of systems have a reference to the debug tools and push errors rather than the tooling pull them. That would at least make it fail during DI injection sooner in the app's runtime. There's probably some cons to it though, for now I'll change it to casting and give a bigger change more thought down the road.

thegoldenmule: In theory I agree but pragmatically it's a problem we will likely never run into. The reason this interface exists is not so much that we can provide other implementations (more sophisticated than passthroughs) but so that we can better test things that depend on it.

PR #255, opened 2019-02-13
Debug Utilities + Crash Logging

thegoldenmule:
* Optimization HUD.
	* Allows toggling assets and scripts by tags.
	* Allows destroying anchors.
* Added capture tab to performance menu.
	* Start/stop.
	* Dumps are sent via email.
* Consolidated debugging tools into DebugService.
* Added a new CrashService that helps with crashes that are not handled by CoreApplication (in UwpCrashLogger).


thegoldenmule:
One thing I don't like: this is a different approach from `DebugService`. I kinda think that either all of the debug tools should go into the service, or we should pull the logging out of the service and get rid of it. I'm thinking the former would be cleaner.

Assets/Source/Application/States/Play/Hmd/IPrimaryAnchorManager.cs
```diff
@@ -28,10 +45,15 @@ public interface IPrimaryAnchorManager
...
-        /// Called to get reference to primary anchor
+        /// Modifies a position/rotation relative to a located anchor. The primary anchor is prioritized.
+        /// The anchor used for relative positioning is returned. If all anchors aren't located, null is returned.
+        /// TODO: Remove UnityEngine dependencies
         /// </summary>
-        WorldAnchorWidget Anchor { get; }
+        /// <param name="position"></param>
+        /// <param name="rotation"></param>
+        /// <returns></returns>
+        WorldAnchorWidget RelativeTransform(ref Vector3 position, ref Quaternion rotation);
```
thegoldenmule: booo, get this outta here

PR #259, opened 2019-02-20
Allow experiences to define custom funtionality for locating prompts

jsundra:
This allows a custom message to be set for locating anchors, as well as whether the anchor bypass button is disabled or not.

The accessors for the scripting API are a bit gross, but it avoids caching the root and dealing with listening for experience loading and whatnot, which seems too heavy for a js api to deal with.

### DIFF
--- a/Assets/Source/Player/Scripting/Interface/App/SystemJsApi.cs
+++ b/Assets/Source/Player/Scripting/Interface/App/SystemJsApi.cs
@@ -82,6 +82,12 @@ namespace CreateAR.EnkluPlayer.Scripting
             get
             {
                 var sceneId = _sceneManager.All.FirstOrDefault();
+                if (string.IsNullOrEmpty(sceneId))
+                {
+                    Log.Error(this, "No scene found.");
+                    return string.Empty;
+                }
+                
                 var root = _sceneManager.Root(sceneId);
                 var schema = root.Schema.Get<string>(PrimaryAnchorManager.PROP_LOCATING_MESSAGE_KEY);
                 return schema.Value;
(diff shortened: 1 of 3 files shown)


Assets/Source/Player/Scripting/Interface/App/SystemJsApi.cs
```diff
@@ -66,6 +74,48 @@ public class SystemJsApi
...
+        /// <summary>
+        /// Allows a custom message to display when the device is locating anchors.
+        /// </summary>
+        public string locatingMessage
+        {
+            get
+            {
+                var sceneId = _sceneManager.All.FirstOrDefault();
+                var root = _sceneManager.Root(sceneId);
+                var schema = root.Schema.Get<string>(PrimaryAnchorManager.PROP_LOCATING_MESSAGE_KEY);
```
thegoldenmule: Null check on root since `FirstOrDefault` can return empty.

Assets/Source/Player/Scripting/Interface/App/SystemJsApi.cs
```diff
@@ -66,6 +74,48 @@ public class SystemJsApi
...
+                return schema.Value;
+            }
+            set 
+            { 
+                var sceneId = _sceneManager.All.FirstOrDefault();
+                var root = _sceneManager.Root(sceneId);
+                var schema = root.Schema.Get<string>(PrimaryAnchorManager.PROP_LOCATING_MESSAGE_KEY);
```
thegoldenmule: same here

Assets/Source/Player/Scripting/Interface/App/SystemJsApi.cs
```diff
@@ -66,6 +74,48 @@ public class SystemJsApi
...
+                var root = _sceneManager.Root(sceneId);
+                var schema = root.Schema.Get<bool>(PrimaryAnchorManager.PROP_DISABLE_BYPASS_KEY);
+                return schema.Value;
+            }
+            set 
+            { 
+                var sceneId = _sceneManager.All.FirstOrDefault();
+                var root = _sceneManager.Root(sceneId);
+                var schema = root.Schema.Get<bool>(PrimaryAnchorManager.PROP_DISABLE_BYPASS_KEY);
+                schema.Value = value;
```
thegoldenmule: same here

PR #262, opened 2019-02-26
IUX Occlusion

jsundra:
Simple MaterialManager for IUX that allows for `ActivatorPrimitive`s and `TextPrimitive`s to change their render type based on a `display` schema prop. Currently supports Overlay and Occluded.

Needs more work down the road. Placeholders for Hidden are there too, but that'll require more work later to feed gestures into material props. In addition, it'd be great to force the Unity renderers to use the different materials and not drive everything thru `.material`. `WidgetRenderer` looks like it can use some cleanup too in how it manages its Material vs the materials actually on its underlying Renderer.

Also, misc - I nuked the Ctrl+A keybind for the application config editor. One too many curses when trying to select all in text fields :)

Example of mixed occlusion + overlay:
iux_occlusion


Assets/Source/Player/IUX/MaterialManager.cs
```diff
@@ -0,0 +1,106 @@
...
+            } 
+            else if (primitive is ActivatorPrimitive)
+            {
+                switch (displayStyle)
+                {
+                    case DisplayStyle.Occluded:
+                        return _config.ButtonOccluded;
+                    default:
+                        return _config.ButtonOverlay;
+                }
```
thegoldenmule: Missing braces throughout.

Assets/Source/Player/IUX/MaterialManager.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        /// <summary>
+        /// Gets a shared Material instance for a primitive based on a string DisplayStyle.
+        /// Parsing errors default to DisplayStyle.Default.
+        /// </summary>
+        /// <param name="primitive">The primitive.</param>
+        /// <param name="displayTypeStr">The display style desired.</param>
+        /// <returns>A Unity material instance.</returns>
+        public Material Material(Element primitive, string displayTypeStr)
```
thegoldenmule: Seems like this function should exist OR the other one. I actually think it might make more sense to not use enums internally to the material manager, but keep them externally for the elements that want to use it. This allows the possibility of expanding this api such that custom materials could be created.

Assets/Source/Player/IUX/MaterialManager.cs
```diff
@@ -0,0 +1,106 @@
+using System;
+using UnityEngine;
+
+namespace CreateAR.EnkluPlayer.IUX
+{
+    /// <summary>
+    /// Manages getting shared Unity Material instances for primitives based on their display style.
+    /// </summary>
+    public class MaterialManager
```
thegoldenmule: There should be an `IMaterialManager` interface defined and used throughout.

Assets/Source/Player/IUX/IMaterialManager.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <returns></returns>
+        Material Material(Element element, string materialStr);
+
+        /// <summary>
+        /// Registered a Material for a given Type.
+        /// </summary>
+        /// <param name="type"></param>
+        /// <param name="material"></param>
+        void RegisterMaterial(Type type, Material material);
+    }
```
thegoldenmule: Niceeeee

PR #264, opened 2019-03-07
Feature/log config

thegoldenmule:
Separate configurations may be provided for each individual log target type.

### DIFF
--- a/Assets/Resources/ApplicationConfig.WSAPlayerX86.json
+++ b/Assets/Resources/ApplicationConfig.WSAPlayerX86.json
@@ -1,5 +1,21 @@
 {
 	"Log": {
-		"Level": "Warning"
-	}
+        "Targets": [
+            {
+                "Target": "unity",
+                "Enabled": false
+            },
+            {
+                "Target": "file",
+                "Level": "Warning",
+                "IncludeObject": true
+            },
+            {
+            	"Target": "loggly",
+            	"Level": "Error",
+            	"IncludeObject": true,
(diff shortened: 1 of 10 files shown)


Assets/Resources/ApplicationConfig.WSAPlayerX86.json
```diff
@@ -1,5 +1,21 @@
 {
 	"Log": {
-		"Level": "Warning"
-	}
+        "Targets": [
+            {
+                "Target": "unity",
+                "Enabled": false
+            },
```
jsundra: By default, this would make it so running in editor with a UWP target would have no logging. Maybe that's okay and the local copy just changes if that's your dev setup? Definitely would be prettier than hard-coding it true if UNITY_EDITOR is defined

thegoldenmule: Hmmm, good point!

thegoldenmule: I've added a new config that is active only in the editor: `ApplicationConfig.Editor.json`. This should fill the need.

Assets/Source/Application/Injection/EnkluPlayerModule.cs
```diff
@@ -775,5 +725,87 @@ private void SetupLogging(InjectionBinder binder, LogAppConfig config)
...
+                var target = Target(binder, targetConfig);
+                if (null != target)
+                {
+                    Log.AddLogTarget(target);
+                }
+            }
+        }
+
+        /// <summary>
+        /// /Creates a log target from a target config,
```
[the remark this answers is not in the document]
thegoldenmule: Fixed,

PR #268, opened 2019-03-19
Add support for selecting video capture's audio source via scripting

jsundra:
Title says it all. In an effort to create less UnityEngine dependencies, I created a `VideoAudioSource` next to the interface, which mirrors Unity's audio enum. Not entirely sure if I like it. Thoughts?


Assets/Source/Application/States/Play/Hmd/HoloLensVideoCapture.cs
```diff
@@ -150,7 +151,24 @@ public IAsyncToken<Void> Setup()
...
+                        break;
+                    case VideoAudioSource.Mic:
+                        unityAudioState = VideoCapture.AudioState.MicAudio;    
+                        break;
+                    case VideoAudioSource.Mixed:
+                        unityAudioState = VideoCapture.AudioState.ApplicationAndMicAudio;    
+                        break;
+                    case VideoAudioSource.None:
+                        unityAudioState = VideoCapture.AudioState.None;
+                        break;
```
thegoldenmule: Missing braces.

Assets/Source/Player/Scripting/Interface/App/VideoCaptureJsApi.cs
```diff
@@ -34,18 +35,38 @@ public VideoCaptureJsApi(IVideoCapture videoCapture)
...
+        public void setupSource(Engine engine, JsFunc callback, string audioSourceStr)
+        {
+            var audioSource = VideoAudioSource.Mixed;
+            try
+            {
+                audioSource = (VideoAudioSource) Enum.Parse(typeof(VideoAudioSource), audioSourceStr);
+            }
+            catch (ArgumentException)
+            {
+                Log.Error(this, "Unknown AudioSource ({0})", audioSourceStr);
```
thegoldenmule: Seems like in this case the callback should be called with a failure.

PR #276, opened 2019-04-19
Standalone Player

thegoldenmule:


thegoldenmule:
Includes "support" for kinect and camera type elements.

PR #277, opened 2019-04-29
Multiplayer Reconcile

thegoldenmule:
Improves reconcile for create/delete.

* Every time the client receives an event that an element has been created, it tracks that new element.
  * If we receive a delete event, it deletes the element and stops tracking it.
  * If instead we receive a scene diff, that means we have disconnected and reconnected. In this case, look over the scene diff.
    * If the scene diff contains a create that we are not already tracking, create the element.
    * If the scene diff contains a create that we are already tracking, ignore the create.
    * If the scene diff does not contain a create, then we know the element has been deleted while the client was offline-- delete the element.

### DIFF
--- a/Assets/Source/Network/Multiplayer/UwpTcpConnection.cs
+++ b/Assets/Source/Network/Multiplayer/UwpTcpConnection.cs
@@ -570,6 +570,7 @@ namespace CreateAR.EnkluPlayer
             _socket = new StreamSocket();
             _socket.Control.KeepAlive = true;
             _socket.Control.NoDelay = true;
+            _socket.Control.QualityOfService = SocketQualityOfService.LowLatency;
 
             try
             {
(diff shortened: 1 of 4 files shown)


Assets/Source/Network/Multiplayer/UwpTcpConnection.cs
```diff
@@ -570,6 +570,7 @@ public bool Connect(string host, int port)
             _socket = new StreamSocket();
             _socket.Control.KeepAlive = true;
             _socket.Control.NoDelay = true;
+            _socket.Control.QualityOfService = SocketQualityOfService.LowLatency;
```
mbolt35: Yessssss

thegoldenmule:
I suppose so. Reloading the scene each time we reconnect isn't really an option.

PR #286, opened 2019-05-20
Feature/kinect composite

jsundra:
Bulk of this PR is to add support for Kinect based image compositing for standalone builds. Compositing is configured through scripting with the ability to specify a specific Kinect sensor, or use whatever is available. The Kinect scripting interface is designed to only exist on platforms that support it.

Standalone build changes:
- Adds a message broadcast when the application exits for standalone builds to use.
- Uses the new standalonewindows asset bundles.


Assets/Source/Player/Scripting/Interface/KinectJsInterface.cs
```diff
@@ -0,0 +1,48 @@
+using CreateAR.Commons.Unity.Messaging;
+
+#if UNITY_STANDALONE_WIN
```
thegoldenmule: I think we should actually return this for all platforms and the context can decide where or not to no-op. It sucks to have to do a null check whenever you use this particular require.

Assets/Source/IoT/Kinect/KinectCamera.cs
```diff
@@ -0,0 +1,212 @@
...
+			_material.SetFloat("_Mirror", mirror ? -1 : 1);
+		}
+	
+		/// <summary>
+		/// Unity initialization.
+		/// </summary>
+		private void Awake()
+		{
+			GetComponent<Camera>().depthTextureMode = DepthTextureMode.Depth;
+			_material = new Material(Shader.Find("Enklu/Composite"));
```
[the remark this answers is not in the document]
jsundra: Shaders are kind of their own special quirk - Any "path" isn't tied to OS file system at all, but just whatever is defined in the shader. So as long as it's included anywhere in the build it'll be find-able.

But your point is still entirely valid, and this will just be another use case of "How do we ensure it will definitely be in the build" other than grossly having our own `Resources` folder :/

Assets/Source/Player/Scripting/Interface/KinectJsInterface.cs
jsundra: I'm on the fence about this. I definitely agree that something like a `requires` statement feels like it should always work. But at the same time, I think it's more performant if it doesn't. The current script I'm using to test Kinect functionality is this:
```
const kinect = require('kinect');

if (kinect) {
    var ctx = kinect.context();
    module.exports = {
        enter: function() {
            ctx.startComposite(this);
            ctx.mirrorComposite = true;
        },
        exit: function() {
            ctx.stopComposite();
        }
    }
}
```

The reason I wrote it like this, is so that a HoloLens runtime early-outs as quickly as possible. If this script required an update loop for Kinect and just had a no-op implementation, there'd be a perf hit on HoloLens still just to invoke the update loop and then the passthrough interface's method.

thegoldenmule: That's fine, but I think we can put the same thing in the interface.

```
const kinect = require('kinect');

if (kinect.isSupported) {
...
}
```
```

Assets/Source/IoT/Kinect/KinectCamera.cs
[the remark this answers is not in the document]
thegoldenmule: You're totes right.

PR #287, opened 2019-05-28
Inventory 

mbolt35:
First pass. Still needs testing.


Assets/Source/Player/Inventory/UserInventoryController.cs
```diff
@@ -0,0 +1,118 @@
...
+{
+    /// <summary>
+    /// Controls user inventory loading and interaction.
+    /// </summary>
+    public class UserInventoryController : IUserInventoryController
+    {
+        /// <summary>
+        /// Http Service used to make stargazer calls.
+        /// </summary>
+        private IHttpService _http;
```
thegoldenmule: `readonly`?

Assets/Source/Player/Inventory/IUserInventoryController.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Collections.Generic;
+using CreateAR.Commons.Unity.Async;
+
+namespace CreateAR.EnkluPlayer.Inventory
+{
+    /// <summary>
+    /// Implementation prototype for a controller capable of loading inventory for users.
+    /// </summary>
+    public interface IUserInventoryController
```
thegoldenmule: This interface seems too static to me as there is no way for it to ever update. I suspect we will have to redo this entirely when users can be granted items dynamically. What about something like:

```
ReadOnlyCollection<InventoryItem> All { get; }

IAsyncToken<Void> Load(userId);

void Unload();
```

Then _this_ object (not its callers) would be responsible for making sure inventory were update to date and consistent.

Assets/Source/Player/Inventory/UserInventoryController.cs
mbolt35: It wasn't even needed.

PR #288, opened 2019-05-28
Feature/reactive audio

jsundra:
Targeted against `feature/mesh-trails` because I screwed up the branching.

Opening before the mesh trail PR for early review/feedback.

Adds support for processing audio sources to add some basic reactive functionality based on output level. Processing hooks allow for scripting to bind audio level -> schema values.

Also includes a very basic particle API to allow triggering emission.

Names of everything are kind of bad..... Open to better name suggestions!!

Example usage:
```
this.audio.hangTime = 0.25;
	
this.audio.mapLevelVec3(this, 'scale', -30, -10, vec3(0.2, 0.2, 0.2), vec3(1, 1, 1));

var child = this.children[0];
if (!child.material) {
    log.error('Child has no material!');
    return;
}
child.material.makeUnique();

child.material.setFloat('_Amplitude', 0);
this.audio.mapLevelFloat(child, 'material._Amplitude', -30, 0, 0, 1);

var child2 = this.children[1];
if (!child2.particles) {
    log.error('Child has no particles');
    return;
}

child2.particles.emitOn('emit', 10);
this.audio.mapLevelBool(child2, 'emit', -15);
```


Assets/Source/Player/Audio/AudioManager.cs
```diff
@@ -0,0 +1,46 @@
...
+            _audioSources.Remove(audioSource);
+        }
+
+        /// <summary>
+        /// Starts processing all registered IAudioSources.
+        /// </summary>
+        public void Process()
+        {
+            // TODO: Copy before iterating, in case schema watching triggers a removal?
+            for (var i = 0; i < _audioSources.Count; i++)
```
thegoldenmule: Can `.Process()` call something that would remove the source?

Assets/Source/Player/Scripting/Interface/UnityAudioSource.cs
```diff
@@ -67,14 +157,196 @@ public float DopplerLevel
...
+                MaxDb = maxDb,
+                MappedMin = mappedMin,
+                MappedMax = mappedMax
+            });
+        }
+
+        /// <inheritdoc />
+        public void UnmapLevelFloat(ElementSchemaProp<float> prop)
+        {
+            for (var i = 0; i < _floatMappings.Count; i++)
```
thegoldenmule: Either null check or disallow null with `readonly`. This could be called before `Map`.

Assets/Source/Util/Extensions/FloatExtensions.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+
+namespace CreateAR.EnkluPlayer
+{
+    /// <summary>
+    /// Float Extensions.
+    /// </summary>
+    public static class FloatExtensions
```
thegoldenmule: We've been putting extensions in the default namespace so you don't have to include.

PR #289, opened 2019-05-29
Feature/mesh trails

jsundra:
Opening for visibility. Still playing with some mesh creation perf logic & shader.

Adds a new TrailWidget that's responsible for building a mesh as its local position changes. Limited functionality right now, just a simple width/color. Local position is used intentionally to prevent one of the eyesores of Unity's TrailRenderer, where giant trails would be created when the device loses tracking and anchors jump around.

Lifecycle is intended to be owned by scripting - multiplayer API builds elements, clients decide when they are unused, and change the sync'd `trail.destroy` schema prop to destroy the element.


thegoldenmule:
As discussed, I do not like this as a long-term strategy. However, a builtin mesh trail is definitely handy!

PR #290, opened 2019-05-31
Heartbeat

mbolt35:


Assets/Source/Network/Multiplayer/KeepAliveHandler.cs
```diff
@@ -0,0 +1,180 @@
...
+            var now = DateTimeOffset.UtcNow;
+
+            // Check for send interval
+            if (null == _outboundRequest)
+            {
+                if (now - _lastReceived >= _sendInterval && _isConnected())
+                {
+                    Log.Debug(this, "Sending Ping Request");
+
+                    _outboundRequest = new PingRequest { PingId = ++_pingId };
```
thegoldenmule: I guess this is okay to just overflow? Seems like a mod would be safer.

mbolt35: Yes, it overflows. It won't hurt anything. I believe this code is ok even with "allow unsafe code" disabled. I can't remember exactly, but I believe C# has something like:

```
checked {
    ++_pingId
}
```
For overflow protection. We can mod for clarity though.

mbolt35: Err, actually we're going to run into a similar issue right? This is just designed to increment and wrap. I guess we can add something like:

```
_pingId = _pingId == byte.MAX_VALUE ? 0 : _pingId + 1;
```

thegoldenmule: If you're sure it's safe, I'm happy.

PR #295, opened 2019-07-16
Object Selection Improvements

thegoldenmule:
There is only so much we can do until we rewrite the edit view, however, this gets us some of the way.

- Centralize all collision mesh creation inside of Widget.
- For assets, replace box collider with mesh colliders.
- For multi-mesh assets, use CombineMeshes rather than many meshes (I found some assets have >100 meshes, which CombineMeshes deals with handily, but individual colliders can crash unity).
- Use the emission shape to create a collider for particle systems.
- If there is no asset, there is no collider.


Assets/Source/Application/States/Play/Desktop/DesktopDesignController.cs
```diff
@@ -393,13 +393,19 @@ private void Editor_OnSelectionChanged(ObjectSelectionChangedEventArgs args)
...
                 {
-                    Log.Error(this, "No controller.");
-                    return;
-                }
+                    if (null == gameObject.transform.parent)
+                    {
+                        return;
+                    }
 
+                    gameObject = gameObject.transform.parent.gameObject;
```
jsundra: Minor, this could stay as a `Transform` and not go back and forth to `GameObject` since `GetComponent` exists on both.

Assets/Source/Player/IUX/Content/ContentWidget.cs
```diff
@@ -260,6 +266,182 @@ protected override void UpdateInternal()
...
+        {
+            var meshGameObject = meshFilter.gameObject;
+            var collider = meshGameObject.GetComponent<MeshCollider>();
+
+            // if this mesh already has a collider, skip it-- we can't
+            // add another collider, and we shouldn't mess with the 
+            // collider on the asset
+            if (null != collider)
+            {
+                return;
```
jsundra: Still reading so maybe it'll be clear later - if this is skipped, how do assets with root level mesh colliders ever become selectable in the editor?

Yeah after reading more this isn't immediately jumping out. `SetupColliders` maybe should add a box collider to the `ContentWidget` itself if no colliders can be built from the asset?

Assets/ThirdParty/Runtime Level Design/Scripts/Runtime Package Common/Camera/RTFocusCamera.cs
```diff
@@ -415,6 +415,17 @@ private float CalculateScrollZoomAmount(float deviceScroll)
...
+            // clamp
+            const float max = 2.5f;
+            if (zoomAmount < -max)
+            {
+                zoomAmount = -max;
+            }
+            else if (zoomAmount > max)
+            {
+                zoomAmount = max;
+            }
```
jsundra: Minor, but since I assume this imports `UnityEngine`, you could use `Mathf.Clamp(zoomAmount, -max, max)`

Assets/Source/Player/IUX/Content/ContentWidget.cs
thegoldenmule: That's a good idea. I'll add that.

Assets/ThirdParty/Runtime Level Design/Scripts/Runtime Package Common/Camera/RTFocusCamera.cs
thegoldenmule: _facepalm_

Assets/ThirdParty/Runtime Level Design/Scripts/Selection/RTObjectSelection.cs
```diff
@@ -1120,36 +1130,51 @@ private bool CanSelectObject(GameObject gameObject, SelectRestrictFlags restrict
...
+            IRTObjectSelectionListener selectionListener = elGameObject.GetComponent<IRTObjectSelectionListener>();
             if (selectionListener != null &&
                 !selectionListener.OnCanBeSelected(new ObjectSelectEventArgs(selectReason)))
             {
                 return false;
             }
             
+            selection = elGameObject;
+
+            Log.Info(this, "{0} Clicked on.", selection.name);
```
thegoldenmule: Derp

Assets/Source/Application/States/Play/Desktop/DesktopDesignController.cs
thegoldenmule: I like minor.

PR #298, opened 2019-07-16
Live updates

thegoldenmule:
- Pull schema value from root to determine if we should auto-connect to Trellis.
- Default to true.


Assets/Source/Application/States/Play/App/AppController.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using CreateAR.Commons.Unity.Async;
 using CreateAR.Commons.Unity.Logging;
+using UnityEngine.Assertions.Must;
```
jsundra: Unused import

thegoldenmule: :+1:

PR #301, opened 2019-07-17
Fixed issue where rotation and scale sometimes did not apply

thegoldenmule:
The problem, I believe, was that we were not looking at transform updates explicitly from drag start to end. In my repro case, I was having the issue 100% of the time and this resolves it for me.


Assets/Source/Application/States/Play/Desktop/Controllers/ElementUpdateMonobehaviour.cs
```diff
@@ -117,38 +126,46 @@ public void OnEndDrag()
...
 
             // check for rotation changes
             {
-                if (!transform.localRotation.eulerAngles.Approximately(_rotationProp.Value))
+                if (!endRotation.Approximately(_startRotation))
                 {
-                    _rotationProp.Value = transform.localRotation.eulerAngles.ToVec();
+                    Log.Info(this, "Rotation changed from {0} -> {1}",
+                        _startRotation,
+                        endRotation);
```
jsundra: Might not want to leave this log in? Or maybe it's useful. _shrugs_

thegoldenmule: I'll remove the log.

jsundra:
I guess I could have read the PR desc to answer my first question! Second point seems valid still though - along the lines of what is changing and what is expected to change. If the answer is "they're being set again for good measure" - I'm totally okay with that :)

thegoldenmule:
The main idea here is that we are pulling off the state of the transform before setting any values. Setting values can change the outcome as we're evaluating it.

PR #303, opened 2019-08-02
Hotfix/script sharing

thegoldenmule:
Scripts no longer have a version on them. Instead, a hash is included in the uri. However, the player was still trying to load a version from the record, which resulted in the version always being 0. Instead, the script cache now uses the hash from the URI.


Assets/Source/Application/Injection/EnkluPlayerModule.cs
```diff
@@ -720,19 +720,12 @@ public void Load(InjectionBinder binder)
             {
                 binder.Bind<JavaScriptParser>().ToValue(new JavaScriptParser(false));
                 binder.Bind<IScriptParser>().To<DefaultScriptParser>().ToSingleton();
-
-                if (UnityEngine.Application.isEditor)
```
thegoldenmule: The fact that the editor cache differently than device caused us to miss this bug during testing. I don't see a reason why the caching should be different.

mbolt35: Yikes. Tricky bug. Good stuff.

PR #305, opened 2019-08-07
Feature/edit mode improvements

michaelcvoigt:
Creating a pull request so folks can try this out. Code is still prototype.


Assets/Resources/ApplicationConfig.Editor.json
```diff
@@ -1,15 +1,31 @@
...
                 "IncludeLevel": false,
                 "IncludeTimestamp": false
             },
             {
                 "Target": "file",
                 "Level": "Debug",
+                "Enabled": false,
                 "IncludeObject": true
             }
         ]
```
thegoldenmule: I don't believe this is what we want. This disables all logging when using the editor.

PR #307, opened 2019-08-08
Kinect Depth 2.0!

jsundra:
Now, instead of the scene rendering and the composite happening in post -> The magic happens pre scene rendering by rendering the Kinect's SDF into the camera's framebuffer. This allows for actual depth based compositing, so transparent rendering infront of a real world object won't also display AR content behind bleeding through. This also allows for all shaders / assets to display in the composite feed, and not require special shader passes to be added to non-standard shaders (like vines!).

Note: Blur is not exposed to JavaScript yet, since it has some weird edge cases around the edges when values become infinity/undefined. Needs a little more debugging.
Bonus: Added a depth debug control, exposed all the way up to the JavaScript layer.

image


Assets/Source/Player/IUX/Widget/KinectWidget.cs
```diff
@@ -237,6 +267,13 @@ private void TeardownLocal()
         /// <param name="id">The id of the body.</param>
         private void SetupBody(ulong id)
         {
+            // have we hit the limit already?
+            if (_bodies.Count >= _bodyCount.Value)
+            {
+                Log.Info(this, "Body limit reached, ignoring new body.");
+                return;
+            }
```
jsundra: In the future, we might want to add support for the case where when a tracked body that was counted towards the limit leaves, to re-invoke SetupBody for a tracked, but previously limited.

But for now this is awesomeeeee

thegoldenmule: :+1:

PR #310, opened 2019-08-22
Feature/text effects

jsundra:
Some of this is copied from https://github.com/enklu/create-commons-vine/pull/1, since we're using source files and not a dll from create-commons-vine.

Why's this PR awesome? Vines can now have some effects to boost their contrast. Which should allow for some extra readability in a headset, as well as not having to worry about over exposed or white areas in a Kinect camera use case.

### Vine Parser 
- Added support for `rgb(r, g, b)` and `rgba(r, g, b, a)` to create `Col4` instances.
- Added support for using double quotes to create string literals.

### TextWidget / Text Primitive
- Added support for `textColor`.
- Added support for effects! Current options are `Shadow` and `Outline`. These are configurable through `effect`, and supported with `effect.color` and `effect.offset`.

### Misc
- Fixed an edge case schema bug. Using `GetOwn` with a custom default would return the type's default instead if there was already an existing prop of a different type.
- Fixed a bug where `_propDisplay` leaked a subscription.

`color` is on `TextWidget` since `Widget` upstream calls `GetOwn`. Not the prettiest, since I feel like that property should be on `TextPrimitive` with the other visual related stuffs?

Example usage:
```
<?Vine>

<Text
    label='2'
    fontSize=200
    color=rgb(0, 1, 0)
    effect='Shadow'
    effect.color=rgb(0, 0, 0)
    effect.offset=(5.0, -5.0, 0)
/>
```

image


jsundra:
Annnnnnd forgot to make the Unit test before opening the PR....

Assets/Source/Player/IUX/Renderer/TextRenderer.cs
```diff
@@ -138,5 +205,46 @@ private void RefreshAlignment()
...
+                    _shadow.effectDistance = _effectOffset.ToVector();
+                    break;
+                case TextEffectType.Outline:
+                    _outline = Text.gameObject.AddOrGetComponent<Outline>();
+                    _outline.effectColor = _effectColor.ToColor();
+                    _outline.effectDistance = _effectOffset.ToVector();
+                    break;
+                default:
+                    Log.Warning(this, "Unhandled TextEffectType : {0}", _effectType);
+                    break;
```
thegoldenmule: Missing braces.

thegoldenmule:
Oh also! Let's use `text.shadow` like CSS.

thegoldenmule:
I'm not sure I understand? Could you write out the proposed vine and your issue with it?

jsundra:
@thegoldenmule Sure! So easily we could handle something like...
```
text.shadow = rgb(0, 0, 0)
text.shadow.offset = (5.0, -5.0, 0)
```
orrrr
```
text.shadow = (5.0, -5.0, 0)
text.shadow.color = rgb(0, 0, 0)
```

but I don't think there's a clean way to grab both a Vec & Col without some serious changes to nicely support something that closely resemble the CSS spec like:
```
text.shadow = rgb(0, 0, 0) (5.0, -5.0, 0)
```

thegoldenmule:
Ah I see what you mean. I think we should stick as close as possible to the spec. I have a short term and long term idea:

Short term: How about we limit support to 1 shadow (the spec allows many) with the following props:

`text.shadow.offset`
`text.shadow.blurRadius` -- This is just part of the spec, we could ignore for now.
`text.shadow.color`

Then `text.shadow` is not recognized for anything.

Long term:

We build a reusable parser that runs at the element level, rather than as part of the vine parser. So, for example, `TextWidget` would do something like:

```
var values = _statementParser.Parse(
    Schema.GetOwn('text.shadow').Value,
    "%f %f %f %c");
```

This would parse according to a format, in this case "float float float color".

jsundra:
@thegoldenmule Updated to work off of new props. Example Vines:
```
<Text
  text.shadow.color=rgb(0, 0, 1)
  text.shadow.offset=(2, 2, 0)
/>
```
```
<Text
  text.outline.color=rgb(1, 0, 0)
  text.outline.offset=(2, -2, 0)
/>
```

thegoldenmule:
Remember to add docs here: https://github.com/enklu/enkluplayer/blob/master/docs/element.schema.properties.md

jsundra:
d'oh

thegoldenmule:
@jsundra Looks good to me! Merge it!

PR #314, opened 2019-09-09
Add timeout for mycelium connect

jsundra:
Pretty sure this used to work without this.....?

Didn't look through the history of `UwpHttpService` and other related things to see where the fallback used to come from. But for now - this does the trick. 

Tested on device with/without internet.


Assets/Source/Application/States/Play/Multiplayer/MyceliumMultiplayerController.cs
```diff
@@ -671,6 +671,25 @@ private void GetMultiplayerToken()
 
                     _connect.Succeed(Void.Instance);
                 });
+
+            // TODO: Remove when API call fails due to timeout.
+            _bootstrapper.BootstrapCoroutine(Wait(_config.Network.DiskFallbackSecs, () =>
```
thegoldenmule: I'm _almost_ positive this is not necessary. The actual socket connect we make synchronously with a short timeout. We should test this more thoroughly.

thegoldenmule: @jsundra `IHttpService` has a global timeout.

thegoldenmule:
This is also in https://github.com/enklu/enkluplayer/pull/315.

PR #315, opened 2019-09-09
TrailWidget watches props

jsundra:
Fixes a bug where watched props were never unwatched. Listens for updates to more props, so that offline fallback creation thru the `AppElementJsApi` can set schema values to match visuals with the MP builder.


Assets/Source/Application/States/Play/Multiplayer/MyceliumMultiplayerController.cs
```diff
@@ -671,6 +671,25 @@ private void GetMultiplayerToken()
 
                     _connect.Succeed(Void.Instance);
                 });
+
+            // TODO: Remove when API call fails due to timeout.
+            _bootstrapper.BootstrapCoroutine(Wait(_config.Network.DiskFallbackSecs, () =>
```
thegoldenmule: As stated in the other PR, this should not be necessary. If it is, there is a deeper logic issue as our connection is synchronous.

Assets/Source/Player/IUX/Widget/TrailWidget.cs
```diff
@@ -49,31 +55,38 @@ protected override void LoadInternalBeforeChildren()
...
-            destroyProp.OnChanged += Destroy_OnChanged;
-
+            var segmentsProp = Schema.GetOwn("trail.segments", 4);
+            
+            _widthProp = Schema.GetOwn("trail.width", 0.1f);
+            _colorProp = Schema.GetOwn("trail.color", Col4.White);
+            _distThreshProp = Schema.GetOwn("trail.distThreshold", 0.1f);
+            _destroyProp = Schema.GetOwn("trail.destroy", false);
             _decayTimeProp = Schema.GetOwn("trail.decayTime", 2f);
+            _positionProp = Schema.GetOwn("position", Vec3.Zero);
```
thegoldenmule: https://github.com/enklu/enkluplayer/blob/master/docs/element.schema.properties.md

thegoldenmule:
@jsundra `IHttpService` has a global timeout.

Assets/Source/Application/States/Play/Multiplayer/MyceliumMultiplayerController.cs
mbolt35: Yeah, I was about to ask about this -- This happens synchronously before connection via socket. What does this "fix"?

thegoldenmule: Yah it's not a TCP connection issue, the HTTP request to Trellis to get a multiplayer token needs to time out.

PR #316, opened 2019-09-10
Added camera js api to adjust clip planes.

thegoldenmule:
Allow users to adjust camera clipping planes from scripts.


thegoldenmule:
Changing to develop-- this isn't really a hotfix.

PR #319, opened 2019-09-17
Cleanup

thegoldenmule:
- Moved some Async and Http extensions to those projects.
- Removed usage of MutableAsyncToken. <-- This is the most intense change. It _appears_ that scene loading thrashes around because of the tricky async nature of content loading.
- Deleted tons of unused classes.


Assets/Editor/Tools/Application/EditorBootstrapper.cs
```diff
@@ -26,6 +28,25 @@ public void BootstrapCoroutine(IEnumerator coroutine)
             _coroutines.Add(coroutine);
         }
 
+        /// <inheritdoc />
+        public void Wait(float milliseconds, Action callback)
```
thegoldenmule: This was added to `IBootstrapper` which is part of the Http library.

Assets/Source/Async/Async.cs
```diff
@@ -32,172 +32,6 @@ public TimedOutException() : base("Operation timed out.")
             }
         }
 
-        /// <summary>
```
thegoldenmule: Now in the Async library.

Assets/Source/Player/Element/Touch/TouchManager.cs
```diff
@@ -360,7 +360,13 @@ private void TeardownElement(Element element)
         /// <param name="content">The content.</param>
         private void SetupContent(ContentWidget content)
         {
-            content.OnLoaded.OnSuccess(Content_OnAssetLoaded);
+            content.OnContentLoaded -= Content_OnAssetLoaded;
+            content.OnContentLoaded += Content_OnAssetLoaded;
+
+            if (content.IsContentLoaded)
```
thegoldenmule: This is really the only place we were counting on that mutable async token.

Assets/Source/Player/IUX/Content/ContentWidget.cs
```diff
@@ -514,12 +474,21 @@ private void UpdateAsset()
...
             }
-
+            
             LogVerbose("Refresh asset for {0}.", Id);
 
             var assetId = _srcAssetProp.Value;
             var version = -1;
 
+            // there may not be a valid asset id (36 characters for a GUID + optional version)
+            if (assetId.Length < 36)
```
thegoldenmule: We were doing TONS of stuff (creating tokens and arrays just to immediately resolve them) even if an asset was not assigned to an element.

Assets/Source/Player/IUX/Content/ContentWidget.cs
```diff
@@ -545,6 +514,10 @@ private void UpdateAsset()
         private string[] GetScriptIds()
         {
             var scriptsSrc = _scriptsProp.Value;
+            if (scriptsSrc.Length <= 2)
```
thegoldenmule: Early out for scripts not being assigned. The 2 comes from the fact that for now, if you added then removed a script, the prop value will be `[]`.

Assets/Source/Player/IUX/Content/ContentWidget.cs
michaelcvoigt: @thegoldenmule - Do we use Yoda conditions for these type of IF statements also?

Assets/Source/Player/IUX/Content/ContentWidget.cs
jsundra: Can this comment be added as a comment in code? I know I'll scratch my head a month from now when I look at this :)

Assets/Source/Player/IUX/Content/ContentWidget.cs
```diff
@@ -667,9 +639,10 @@ private void PollScriptLoads()
 
             LogVerbose("All scripts loaded. Running!");
 
-            _onScriptsLoaded.Succeed(this);
-
             _runner.Setup(_enkluScripts);
+
+            IsContentLoaded = true;
```
jsundra: Should `IsContentLoaded` be set back to false when `UpdateAsset` is called again?

thegoldenmule: :+1:

Assets/Source/Player/IUX/Content/ContentWidget.cs
thegoldenmule: :+1:

Assets/Source/Player/IUX/Content/ContentWidget.cs
thegoldenmule: Just for `==`. And I don't think you _have_ to use yogish.

PR #321, opened 2019-09-23
Fix issue with anchor upload where default server body parser cannot …

thegoldenmule:
…handle missing quotes.


Assets/Plugins/CreateAR/Commons/UwpHttpService.cs
```diff
@@ -198,14 +199,33 @@ public IAsyncToken<HttpResponse<byte[]>> Download(string url)
             ref byte[] file)
         {
             var request = new HttpRequestMessage(MethodFor(verb), Services.Urls.Url(url));
-            var form = new MultipartFormDataContent();
+
+            const string BOUNDARY = "Boundary7MA4YWxkTrZu0gW";
```
thegoldenmule: We found this to be reliable from the snap controller.

PR #322, opened 2019-09-25
Fix broken HTTP calls

jsundra:
String.Substring throws an exception if the desired length > string length. Who knew!

Also, for some future proofing, added a guard around OnRequest being called in `UwpHttpService`.


Assets/Source/Application/Services/DebugService.cs
```diff
@@ -630,7 +630,9 @@ private void App_OnLoad(object _)
                     // ignore
                 }
 
-                message += string.Format("\\\n\t-d '{0}'", stringPayload).Substring(0, 250);
+                if (stringPayload.Length > 250) stringPayload = stringPayload.Substring(0, 250);
```
thegoldenmule: Missing braces.

PR #324, opened 2019-10-01
Feature/notification

thegoldenmule:
Support for _receiving_ notifications over mycelium and associated JS API.

```
require('mp')
    .notifications
    .on('type', function(payloadString) {
        // handle
    });
```


thegoldenmule:
We are going to merge this into the Azure Kinect branch.

Assets/Source/Application/States/Play/Multiplayer/MyceliumMultiplayerController.cs
```diff
@@ -584,7 +588,16 @@ private void OnApplicationResume(object obj)
...
         }
-        
+
+        /// <summary>
+        /// Called when there is a notification.
+        /// </summary>
+        /// <param name="evt">The event.</param>
+        private void OnNotificationEvent(NotificationEvent evt)
+        {
+            OnNotification?.Invoke(evt.Type, evt.Payload);
```
jsundra: Might want to try/catch this?

thegoldenmule: I think we can just use normal error propagation here. If this were dispatching into the JS layer, then yes definitely.

PR #325, opened 2019-10-02
Feature/azure kinect

jsundra:
First pass at adding Azure Kinect support, similar to the existing Kinect v2 support. Current support can be used via `require('azure-kinect-preview')`.

Also included is postprocessing and an API to control it.

Kinect Diagram


Assets/Source/Application/States/Play/IVideoUploader.cs
```diff
@@ -0,0 +1,9 @@
+using CreateAR.Commons.Unity.Async;
+
+namespace CreateAR.EnkluPlayer
+{
+    public interface IVideoUploader
+    {
+        IAsyncToken<Void> UploadFile(string fullPath, string orgId, string tag);
```
thegoldenmule: yes please

Assets/AzureTesting/Render.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.IO;
+using UnityEngine;
+
+[RequireComponent(typeof(Camera))]
+public class Render : MonoBehaviour
```
thegoldenmule: We should either delete this (my preference) or bring it up to code standard.

Assets/Resources/Shaders/Composite_KinectV4.shader
```diff
@@ -0,0 +1,102 @@
...
+			    fragOut o;
+			    o.color = float4(r, g, b, 1);
+			    o.depth = outDepth;
+				return o;
+			}
+			
+			//float2 ColorToDepth(float2 coords) {
+			//    coords = (coords * 2 - 1) * _ColorFOV / _DepthFOV;
+			//    return (coords + 1) / 2;
+			//}
```
thegoldenmule: Delete

Assets/Source/Application/Injection/EnkluPlayerModule.cs
```diff
@@ -562,6 +564,14 @@ public void Load(InjectionBinder binder)
 #if NETFX_CORE || (!UNITY_EDITOR && UNITY_WSA)
                 binder.Bind<IImageCapture>().To<HoloLensImageCapture>().ToSingleton();
                 binder.Bind<IVideoCapture>().To<HoloLensVideoCapture>().ToSingleton();
+#elif UNITY_STANDALONE_WIN
+                binder.Bind<IImageCapture>().To<ScreenShotImageCapture>().ToSingleton();
+
+                // Gross? A bit. Need to mull this over.
+                var mainCamera = binder.GetInstance<MainCamera>();
```
thegoldenmule: I'm actually fine with this.

Assets/Source/Application/States/Play/Hmd/HoloLensVideoManager.cs
```diff
@@ -1,19 +1,16 @@
+// TODO: DELETE.
```
thegoldenmule: Do todo.

Assets/Source/Application/States/Play/HoloLensImageManager.cs
```diff
@@ -278,6 +278,7 @@ private async Task UploadFile(string filepath, AsyncToken<Void> token)
 
                 fileStream.Dispose();
                 
+                // TODO: use synchronizer
```
thegoldenmule: Thus could come back on another thread, use `ISynchronizer` to pull it to the main thread to resolve the token.

Assets/Source/Application/States/Play/IVideoCapture.cs
```diff
@@ -68,5 +78,11 @@ public interface IVideoCapture
         /// </summary>
         /// <returns></returns>
         void Teardown();
+
+        /// <summary>
+        /// Sets a custom camera as the video source. May only work on specific platforms.
+        /// </summary>
+        /// <param name="camera"></param>
+        void SetCustomCamera(Camera camera);
```
thegoldenmule: I think `May only work on specific platforms` should be removed. Either the function should be on the interface and implemented by each implementation, or it should be put on the specific implementation that implements it only.

Assets/Source/Application/States/Play/IVideoUploader.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using CreateAR.Commons.Unity.Async;
+
+namespace CreateAR.EnkluPlayer
+{
+    public interface IVideoUploader
+    {
+        IAsyncToken<Void> UploadFile(string fullPath, string orgId, Dictionary<string, string> parameters);
```
thegoldenmule: Comments

Assets/Source/Application/States/Play/PassthroughVideoCapture.cs
```diff
@@ -23,6 +24,12 @@ public IAsyncToken<Void> Setup(VideoAudioSource audioSource)
             return new AsyncToken<Void>(new NotSupportedException());
         }
 
+        /// <inheritdoc />
+        public IAsyncToken<Void> Setup(int width, int height, int framerate)
+        {
+            return new AsyncToken<Void>(new NotSupportedException());
```
thegoldenmule: In the rest of the code base we use "passthrough" to mean silently do nothing-- like a noop. If the implementation throws exceptions it would better be called `ExceptionVideoCapture`.

Assets/Source/Application/States/Play/Standalone/WindowsVideoCapture.cs
```diff
@@ -0,0 +1,189 @@
...
+using System.Collections.Generic;
+using System.IO;
+using CreateAR.Commons.Unity.Async;
+using CreateAR.Commons.Unity.Logging;
+using UnityEngine;
+using Void = CreateAR.Commons.Unity.Async.Void;
+
+namespace CreateAR.EnkluPlayer
+{
+    public class WindowsVideoCapture : IVideoCapture
```
thegoldenmule: Comments

Assets/Source/Application/States/Play/Standalone/WindowsVideoCapture.cs
```diff
@@ -0,0 +1,189 @@
...
+        
+        /// <inheritdoc />
+        public Action<string> OnVideoCreated { get; set; }
+        
+        /// <inheritdoc />
+        public bool IsRecording { get; private set; }
+
+        public WindowsVideoCapture(WindowsVideoCaptureBootstrapper bootstrapper, SnapConfig config)
+        {
+            _bootstrapper = bootstrapper;
```
thegoldenmule: I'm confused by the bootstrappers role, and why it is passed in AND created,

Assets/Source/Application/States/Play/Standalone/WindowsVideoCapture.cs
```diff
@@ -0,0 +1,189 @@
...
+
+            // Hook new instance
+            _bootstrapper = camera.gameObject.AddComponent<WindowsVideoCaptureBootstrapper>();
+            _bootstrapper.OnRender += OnRenderImage;
+        }
+
+        /// <inheritdoc />
+        public IAsyncToken<Void> Setup(VideoAudioSource audioSource)
+        {
+            return new AsyncToken<Void>(new NotSupportedException());
```
thegoldenmule: Why??

Assets/Source/Application/States/Play/Standalone/WindowsVideoCaptureBootstrapper.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using UnityEngine;
+
+namespace CreateAR.EnkluPlayer
+{
+    public class WindowsVideoCaptureBootstrapper : MonoBehaviour
```
thegoldenmule: It seems like `WindowsVideoCapture` should just be a `MonoBehaviour`. This seems like a lot of trouble for nothing.

Assets/Source/Application/States/Play/StandardVideoUploader.cs
```diff
@@ -0,0 +1,140 @@
...
+using CreateAR.Commons.Unity.Http;
+using CreateAR.Commons.Unity.Logging;
+using CreateAR.EnkluPlayer.Session;
+using CreateAR.Trellis.Messages.CreateSnap;
+using Newtonsoft.Json;
+using Void = CreateAR.Commons.Unity.Async.Void;
+
+namespace CreateAR.EnkluPlayer
+{
+    public class StandardVideoUploader : IVideoUploader
```
thegoldenmule: Comments throughout.

Assets/Source/Application/States/Play/VideoManager.cs
```diff
@@ -0,0 +1,311 @@
...
+using CreateAR.Commons.Unity.Async;
+using CreateAR.Commons.Unity.Http;
+using CreateAR.Commons.Unity.Logging;
+using Newtonsoft.Json;
+using UnityEngine;
+using Void = CreateAR.Commons.Unity.Async.Void;
+
+namespace CreateAR.EnkluPlayer
+{
+    public class VideoManager : IVideoManager
```
thegoldenmule: Comments throughout.

Assets/Source/Application/States/Play/VideoManager.cs
```diff
@@ -0,0 +1,311 @@
...
+                existingVideos = Directory.GetFiles(config.Snap.VideoFolder, "*.mp4");
+            }
+            catch
+            {
+                existingVideos = new string[0];
+            }
+
+            var retryCount = 0;
+            
+            Log.Info(this, $"Found {existingVideos.Length} existing videos on disk. Uploading!");
```
thegoldenmule: Don't use string interp in logs. Lean on the `Log` implementation to do this so we aren't paying the cost of string formatting when logs are turned off (these are not compiled out).

Assets/Source/Application/States/Play/VideoManager.cs
```diff
@@ -0,0 +1,311 @@
...
+                        {
+                            File = videoFile,
+                            Info = args
+                        });
+                    
+                        Log.Info(this, "Added to queue!");
+                    }
+                    catch (Exception e)
+                    {
+                        Log.Error(this, $"Error reading cached video data: {e}");
```
thegoldenmule: No interp

Assets/Source/Application/States/Play/VideoManager.cs
```diff
@@ -0,0 +1,311 @@
...
+                Log.Info(this, "Upload request came in while awaiting failure retry.");
+                _isWaiting = false;
+            }
+            
+            _isActive = true;
+            GetOrgId()
+                .OnSuccess(_ => ProcessQueues())
+                .OnFailure(e =>
+                {
+                    Log.Warning(this, $"{e}");
```
thegoldenmule: No need for interp

Assets/Source/Application/States/Play/VideoManager.cs
```diff
@@ -0,0 +1,311 @@
...
+            queue.RemoveAt(0);
+            
+            return _uploader
+                .UploadFile(payload.File, _orgId, payload.Info)
+                .OnSuccess(_ =>
+                {
+                    Log.Info(this, "Upload success. Deleting from disk.");
+                    // TODO: Abstract file I/O out of this
+                    File.Delete(payload.File);
+                    File.Delete(GetJsonFilename(payload.File));
```
thegoldenmule: I would recommend putting try/catch around these. The `File` API throws many an error.

Assets/Source/Application/States/Play/VideoManager.cs
```diff
@@ -0,0 +1,311 @@
...
+                })
+                .OnFailure(e =>
+                {
+                    Log.Error(this, $"Failed to upload video: (File={payload.File} Exception={e}");
```
thegoldenmule: No interp.

thegoldenmule: I would also say this should be `Warn` level. I don't need an email because we lose internet connection.

Assets/Source/Application/States/Play/VideoManager.cs
```diff
@@ -0,0 +1,311 @@
...
+            var path = Path.GetDirectoryName(videoFile);
+            var filename = Path.GetFileNameWithoutExtension(videoFile);
+            return $"{path}/{filename}.json";
+        }
+
+        private void DeleteExcessFile(string filename, string jsonfile)
+        {
+            Log.Info(this, $"Too many files on disk. Removing {filename}");
+            File.Delete(filename);
+            File.Delete(jsonfile);
```
thegoldenmule: Try/catches here.

Assets/Source/IoT/AzureKinect/BodyData.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+using CreateAR.EnkluPlayer.IoT.Kinect;
+using Microsoft.Azure.Kinect.BodyTracking;
+
+namespace CreateAR.EnkluPlayer.IoT.AzureKinect
```
thegoldenmule: Comments.

Assets/Source/IoT/AzureKinect/BodyManagement.cs
```diff
@@ -0,0 +1,213 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using CreateAR.Commons.Unity.Logging;
+using CreateAR.EnkluPlayer.IUX;
+using Enklu.Data;
+
+namespace CreateAR.EnkluPlayer.IoT.AzureKinect
+{
+    public class BodyManagement : IBodyManagement
```
thegoldenmule: Comments.

Assets/Source/IoT/AzureKinect/BodyManagement.cs
```diff
@@ -0,0 +1,213 @@
...
+            var values = _bodyMap.Values.ToList();
+            for (int i = 0, len = values.Count; i < len; i++)
+            {
+                CleanupRecord(values[i]);
+            }
+            
+            _bodyMap.Clear();
+        }
+
+        public void SetAssets(Dictionary<string, string> map)
```
thegoldenmule: I'm not a big fan of set-only functions. Why not just a property?

Assets/Source/IoT/AzureKinect/BodyManagement.cs
```diff
@@ -0,0 +1,213 @@
...
+        {
+            _assetMap = map;
+        }
+
+        private void Body_OnDetected(BodyData data)
+        {
+            Log.Info(this, $"Body detected: {data.BodyId}");
```
thegoldenmule: No interp.

Assets/Source/IoT/AzureKinect/BodyManagement.cs
```diff
@@ -0,0 +1,213 @@
...
+            }
+
+            var record = CreateLocal(data.BodyId);
+            _bodyMap[data.BodyId] = record;
+
+            try
+            {
+                OnBodyDetected?.Invoke(record.Root);
+            }
+            catch (Exception e)
```
thegoldenmule: Please don't use single letter variable names. It's not too much work to type `exception`. :)

Assets/Source/IoT/AzureKinect/BodyManagement.cs
```diff
@@ -0,0 +1,213 @@
...
+            {
+                Log.Error(this, $"Error invoking OnBodyDetected: {e}");
```
thegoldenmule: No interp.

Assets/Source/IoT/AzureKinect/BodyManagement.cs
```diff
@@ -0,0 +1,213 @@
...
+                record.UpdateJoint(kvp.Key.ToString().ToLower().Replace("_", ""), transform.Position, transform.Rotation);
+            }
+            
+            try
+            {
+                OnBodyUpdate?.Invoke(record.Root);
+            }
+            catch (Exception e)
+            {
+                Log.Error(this, $"Error invoking OnBodyUpdate: {e}");
```
thegoldenmule: No interp.

Assets/Source/IoT/AzureKinect/BodyManagement.cs
```diff
@@ -0,0 +1,213 @@
...
+            }
+        }
+
+        private void Body_OnLost(uint id)
+        {
+            Log.Info(this, $"Body lost: {id}");
```
thegoldenmule: Interp

Assets/Source/IoT/AzureKinect/BodyManagement.cs
```diff
@@ -0,0 +1,213 @@
...
+
+        private void CleanupRecord(BodyRecord record)
+        {
+            try
+            {
+                OnBodyLost?.Invoke(record.Root);
+            }
+            catch (Exception e)
+            {
+                Log.Error(this, $"Error invoking OnBodyLost: {e}");
```
thegoldenmule: Interp, no single letter vars.

Assets/Source/IoT/AzureKinect/BodyProcessing.cs
```diff
@@ -0,0 +1,241 @@
...
+using CreateAR.Commons.Unity.Logging;
+using CreateAR.EnkluPlayer.IoT.Kinect;
+using Enklu.Data;
+using Microsoft.Azure.Kinect.BodyTracking;
+using Microsoft.Azure.Kinect.Sensor;
+using UnityEngine;
+
+namespace CreateAR.EnkluPlayer.IoT.AzureKinect
+{
+    public class BodyProcessing : IBodyProcessing
```
thegoldenmule: Comments throughout.

Assets/Source/IoT/AzureKinect/BodyProcessing.cs
```diff
@@ -0,0 +1,241 @@
...
+        
+        public BodyProcessing(CaptureProcessing captureProcessing, IBootstrapper bootstrapper)
+        {
+            _captureProcessing = captureProcessing;
+            _bootstrapper = bootstrapper;
+        }
+
+        public void Setup(TrackerConfiguration configuration = default)
+        {
+            Log.Info(this, $"Setup. ({configuration.SensorOrientation})");
```
thegoldenmule: No interp.

Assets/Source/IoT/AzureKinect/BodyProcessing.cs
```diff
@@ -0,0 +1,241 @@
...
+            }
+            
+            _bodyTracker = new BodyTracker(_captureProcessing.Calibration.Value, configuration);
+            for (var i = 0; i < _bodyData.Length; i++)
+            {
+                var body = new BodyData();
+                for (int j = 0, len = _trackedJoints.Count; j < len; j++)
+                {
+                    body.Joints[_trackedJoints[j]] = new KinectJointTransform();
+                }
```
thegoldenmule: I feel like this could be simplified. Something like `new KinectJointTransforms(_trackedJoints)`

Assets/Source/IoT/AzureKinect/BodyProcessing.cs
```diff
@@ -0,0 +1,241 @@
...
+            _isActive = false;
+
+            try
+            {
+                _bodyTracker.Shutdown();
+                _bodyTracker.Dispose();
+            }
+            catch (Exception e)
+            {
+                Log.Error(this, $"Error closing BodyTracker: {e}");
```
thegoldenmule: No interp, no `e`.

Assets/Source/IoT/AzureKinect/BodyProcessing.cs
```diff
@@ -0,0 +1,241 @@
...
+            }
+        }
+
+        public void ProcessCapture(Capture capture)
```
thegoldenmule: I'm a big fan of simplifying method names when it's clear what they do. `ProcessCapture(capture)` is redundant. `Process(capture)` makes more sense.

Assets/Source/IoT/AzureKinect/BodyProcessing.cs
```diff
@@ -0,0 +1,241 @@
...
+                {
+                    if (_detectedQueue.TryDequeue(out var bodyData))
+                    {
+                        try
+                        {
+                            OnDetected?.Invoke(bodyData);
+                        }
+                        catch (Exception e)
+                        {
+                            Log.Error(this, $"Error invoking body detected: {e}");
```
thegoldenmule: No interp, no `e`.

Assets/Source/IoT/AzureKinect/BodyProcessing.cs
```diff
@@ -0,0 +1,241 @@
...
+                {
+                    if (_updateQueue.TryDequeue(out var bodyData))
+                    {
+                        try
+                        {
+                            OnUpdate?.Invoke(bodyData);
+                        }
+                        catch (Exception e)
+                        {
+                            Log.Error(this, $"Error invoking body update: {e}");
```
thegoldenmule: No interp, no `e`.

Assets/Source/IoT/AzureKinect/BodyProcessing.cs
```diff
@@ -0,0 +1,241 @@
...
+                {
+                    if (_lostQueue.TryDequeue(out var bodyId))
+                    {
+                        try
+                        {
+                            OnLost?.Invoke(bodyId);
+                        }
+                        catch (Exception e)
+                        {
+                            Log.Error(this, $"Error invoking body lost: {e}");
```
thegoldenmule: No interp, no `e`.

Assets/Source/IoT/AzureKinect/BodyProcessing.cs
```diff
@@ -0,0 +1,241 @@
...
+            _scratch.Clear();
+            
+            frame.Dispose();
+        }
+
+        private IEnumerator NotificationLoop()
+        {
+            while (_isActive)
+            {
+                while (_detectedQueue.Count > 0)
```
thegoldenmule: I'm wondering if it would be preferable to have 1 queue that is ordered with ALL such events. This requires all adds, then all updates, then all deletes.  I think that's always safe with respect to the _same_ bodyData, but there may be use cases that require knowledge of, say, how many bodies are in a scene, which may be wrong if a delete actually came before a detected but the detected is dispatched first.

Assets/Source/IoT/AzureKinect/BodyProcessing.cs
```diff
@@ -0,0 +1,241 @@
...
+                            Log.Error(this, $"Error invoking body lost: {e}");
+                        }
+                    }
+                }
+                
+                yield return null;
+            }
+        }
+        
+        private readonly Quaternion TRANSFORM = Quaternion.FromToRotation(Vector3.up, Vector3.forward) * Quaternion.FromToRotation(Vector3.right, Vector3.up);
```
thegoldenmule: Should be at the top of the file.

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+using System.Threading.Tasks;
+using CreateAR.Commons.Unity.Http;
+using CreateAR.Commons.Unity.Logging;
+using Microsoft.Azure.Kinect.BodyTracking;
+using Microsoft.Azure.Kinect.Sensor;
+using UnityEngine;
+
+namespace CreateAR.EnkluPlayer.IoT.AzureKinect
+{
+    public class CaptureProcessing
```
thegoldenmule: Missing comments throughout.

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+    {
+        public enum ProcessingMode
+        {
+            Image,
+            Body,
+            Combined
+        }
+        
+        private Device _kinect;
```
thegoldenmule: Watch your variable ordering. Static then const then readonly then mutable.

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        private readonly ImageProcessing _imageProcessing;
+        private readonly BodyProcessing _bodyProcessing;
+        
+        private CancellationTokenSource _cancellation;
+
+        private ProcessingMode _mode;
+
+        private const int TIME_SAMPLES = 10;
```
thegoldenmule: Is there any reason we'd want to make this configurable?

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+        public Calibration? Calibration { get; private set; }
+        
+        public IImageProcessing ImageProcessing => _imageProcessing;
+
+        public IBodyProcessing BodyProcessing => _bodyProcessing;
+
+        public CaptureProcessing(IBootstrapper _bootstrapper)
+        {
+            _imageProcessing = new ImageProcessing(this);
+            _bodyProcessing = new BodyProcessing(this, _bootstrapper);
```
thegoldenmule: It's unusual, I shall say, for a class to use an interface yet instantiate a concrete implementation. I found it helpful to have a `PassthroughBodyProcessing` implementation for testing. Perhaps we should instead pass a `IProcessingFactory` in with `Image()` and `Body()` methods.

thegoldenmule:
```
public interface IProcessingFactory
{
IImageProcessing Image();
IBodyProcessing Body();
}
```

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+                }
+                return (float) avg / TIME_SAMPLES;
+            }
+        }
+
+        public float TemperatureFahrenheit => _tempCelcius * 1.8f + 32;
+
+        public Calibration? Calibration { get; private set; }
+        
+        public IImageProcessing ImageProcessing => _imageProcessing;
```
thegoldenmule: What do you think about `IImageProcessor` and `IBodyProcessor`? Gerunds in an interface name? What's next, omitting Harvard commas?

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+                return;
+            }
+
+            Calibration = _kinect.GetCalibration();
+                
+            var colorCal = Calibration.Value.ColorCameraCalibration;
+            var depthCal = Calibration.Value.DepthCameraCalibration;
+            Log.Info(this, 
+                $"Video feeds started. (RGB={colorCal.ResolutionWidth}x{colorCal.ResolutionHeight} " +
+                $"D={depthCal.ResolutionWidth}x{depthCal.ResolutionHeight})");
```
thegoldenmule: No interp.

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+
+
+            switch (_mode)
+            {
+                case ProcessingMode.Image:
```
thegoldenmule: MISSING BRACES YOU PHILISTINE!

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+                    _imageProcessing.Setup();
+                    break;
+                case ProcessingMode.Body:
+                    _bodyProcessing.Setup();
+                    break;
+                case ProcessingMode.Combined:
```
thegoldenmule: MISSING BRACES YOU PHILISTINE!

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+                $"Video feeds started. (RGB={colorCal.ResolutionWidth}x{colorCal.ResolutionHeight} " +
+                $"D={depthCal.ResolutionWidth}x{depthCal.ResolutionHeight})");
+
+
+            switch (_mode)
+            {
+                case ProcessingMode.Image:
+                    _imageProcessing.Setup();
+                    break;
+                case ProcessingMode.Body:
```
thegoldenmule: MISSING BRACES YOU PHILISTINE!

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+                    _bodyProcessing.Setup();
+                    break;
+                case ProcessingMode.Combined:
+                    _imageProcessing.Setup();
+                    _bodyProcessing.Setup();
+                    break;
+            }
+            
+            _cancellation = new CancellationTokenSource();
+            Task.Run((Action) ProcessingThread);
```
thegoldenmule: You should use `ThreadHelper.SyncStart`. This abstracts how threads are started (which probably doesn't really matter here) AND configures a `Thread` properly. Tasks are made for short-running tasks-- the pool expects the task to be back shortly. Tasks can be configured to be long running, but I don't think you want to have to worry about it here. Anyway, `ThreadHelper`.

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        public void Teardown()
+        {
+            _cancellation.Cancel();
+
+            switch (_mode)
+            {
+                case ProcessingMode.Image:
```
thegoldenmule: Braces

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+                catch (Exception e)
+                {
+                    Log.Error(this, $"Error getting capture: {e}");
+                    continue;
+                }
+
+                try
+                {
+                    _imageProcessing.ProcessCapture(capture);
+                    _bodyProcessing.ProcessCapture(capture);
```
thegoldenmule: Should these be in separate try/catches? An exception in image processing would kill body processing unnecessarily.

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+                {
+                    Log.Error(this, $"Error processing capture: {e}");
+                }
+
+                _tempCelcius = capture.Temperature;
+                
+                capture.Dispose();
+
+                _durationIndex = ++_durationIndex % TIME_SAMPLES;
+                _durations[_durationIndex] = (DateTime.Now - now).TotalMilliseconds;
```
thegoldenmule: This is fine, but `Stopwatch` may be a better class to use here.

Assets/Source/IoT/AzureKinect/CaptureProcessing.cs
```diff
@@ -0,0 +1,201 @@
...
+            
+            Log.Info(this, "Processing thread spun down.");
+        }
+
+        private DeviceConfiguration BuildConfig(ProcessingMode mode)
+        {
+            // TODO: Make camera mode configurable.
+            switch (mode)
+            {
+                case ProcessingMode.Image:
```
thegoldenmule: Missing braces.

thegoldenmule: Also: can these be more configurable?

Assets/Source/IoT/AzureKinect/IBodyProcessing.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace CreateAR.EnkluPlayer.IoT.AzureKinect
+{
+    public interface IBodyProcessing
```
thegoldenmule: Missing braces. Also missing several functions.

Assets/Source/IoT/AzureKinect/IImageProcessing.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace CreateAR.EnkluPlayer.IoT.AzureKinect
+{
+    public interface IImageProcessing
```
thegoldenmule: Same here, this is actually missing some functions.

Assets/Source/IoT/AzureKinect/IImageProcessing.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public IRawImage Color;
+        public IRawImage Depth;
+    }
+
+    public interface IRawImage : IDisposable
+    {
+        IntPtr Address { get; }
+        int Length { get; }
+    }
```
thegoldenmule: <nitpick>The struct and interface should go above `IImageProcessing`.</nitpick>

Assets/Source/IoT/AzureKinect/IKinectContext.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using CreateAR.EnkluPlayer.IUX;
+using CreateAR.EnkluPlayer.Scripting;
+
+namespace CreateAR.EnkluPlayer.IoT.AzureKinect
+{
+    public interface IKinectContext : IKinectCommonContext
```
thegoldenmule: Hmmmm, I need to read more on the strategy here of extending the interface.

Assets/Source/IoT/AzureKinect/ImageProcessing.cs
```diff
@@ -0,0 +1,107 @@
+using System;
+using System.Buffers;
+using System.Collections.Concurrent;
+using CreateAR.Commons.Unity.Logging;
+using Microsoft.Azure.Kinect.Sensor;
+
+namespace CreateAR.EnkluPlayer.IoT.AzureKinect
+{
+    public class ImageProcessing : IImageProcessing
```
thegoldenmule: Comments throughout.

Assets/Source/IoT/AzureKinect/ImageProcessing.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        private class RawImage : IRawImage
```
thegoldenmule: Is there a reason to use an interface here? I'm not seeing the benefit of other implementations here. I think it might make more sense to just use `RawImage` directly in the `IImageProcessing` interface.

Assets/Source/IoT/AzureKinect/ImageProcessing.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        private readonly CaptureProcessing _captureProcessing;
+        private Transformation _transformation;
+
+        private readonly ConcurrentQueue<Images> _queue = new ConcurrentQueue<Images>();
+
+        private bool _isActive;
+
+        public ImageProcessing(CaptureProcessing captureProcessing)
```
thegoldenmule: I'm thinking that whatever parts of `CaptureProcessing` this is using should probably be on an interface that `CaptureProcessing` uses or a POCO that `CaptureProcessing` can pass to it. Having the bi-directional dependency here is giving me pause (`CaptureProcessing` creates this object and passes itself to it).

Assets/Source/IoT/AzureKinect/KinectCamera.cs
```diff
@@ -0,0 +1,92 @@
+using CreateAR.Commons.Unity.Logging;
+using UnityEngine;
+
+namespace CreateAR.EnkluPlayer.IoT.AzureKinect
+{
+    [RequireComponent(typeof(Camera))]
+    public class KinectCamera : InjectableMonoBehaviour
```
thegoldenmule: Missing comments throughout.

Assets/Source/IoT/AzureKinect/KinectCamera.cs
```diff
@@ -0,0 +1,92 @@
...
+    {
+        [Inject]
+        public IVideoCapture _videoCapture { get; set; }
+        
+        private static readonly int _DepthTex = Shader.PropertyToID("_DepthTex");
+        private static readonly int _DebugDepth = Shader.PropertyToID("_DebugDepth");
+        private static readonly int _MaxDepth = Shader.PropertyToID("_MaxDepth");
+        
+        // Serialize these fields to allow unity inspector debugging
+        [SerializeField] private Material _material;
```
thegoldenmule: `[SerializeField]` should be on separate line.

Assets/Source/IoT/AzureKinect/KinectCamera.cs
```diff
@@ -0,0 +1,92 @@
...
+            get { return _material.GetFloat(_DebugDepth) > 0.5f; }
+            set { _material.SetFloat(_DebugDepth, value ? 1 : 0); }
+        }
+
+        protected override void Awake()
+        {
+            Log.Info(this, "Building camera.");
+            base.Awake();
+            
+            _material = new Material(Shader.Find("Enklu/Composite_KinectV4"));
```
thegoldenmule: Idea: we have `Shader.Find` calls in a growing number of places. Perhaps we should have a public static `EnkluShaders` file that does the lookups for us.

Assets/Source/IoT/AzureKinect/KinectContext.cs
```diff
@@ -0,0 +1,282 @@
...
+using CreateAR.Commons.Unity.Messaging;
+using CreateAR.EnkluPlayer.IUX;
+using CreateAR.EnkluPlayer.Scripting;
+using Microsoft.Azure.Kinect.Sensor;
+using UnityEngine;
+using UnityEngine.Rendering.PostProcessing;
+
+namespace CreateAR.EnkluPlayer.IoT.AzureKinect
+{
+    public class KinectContext : IKinectContext
```
thegoldenmule: Missing comments throughout.

Assets/Source/IoT/AzureKinect/KinectContext.cs
```diff
@@ -0,0 +1,282 @@
...
+    {
+        private const string METRICS_PREFIX = "AzureKinect";
+        
+        private readonly IBootstrapper _bootstrapper;
+        private readonly IMetricsService _metrics;
+        private readonly Camera _mainCamera;
+        private readonly PostProcessProfile _profile;
+        private readonly PostProcessResources _resources;
+
+        private Device _kinect;
```
thegoldenmule: Variable ordering: this should be below `readonly`s.

Assets/Source/IoT/AzureKinect/KinectContext.cs
```diff
@@ -0,0 +1,282 @@
...
+        public void Start(ContentWidgetJs element)
+        {
+            if (element != null)
+            {
+                var parent = element.ContentWidget.GameObject;
+                _kinectCamera = parent.AddOrGetComponent<KinectCamera>();
+                _kinectCamera.MaxDepth = _maxDepth;
+                _kinectCamera.DebugDepth = _debugDepth;
+
+                if (_profile)
```
thegoldenmule: Gah, we need to make a real system here.

Assets/Source/IoT/AzureKinect/KinectContext.cs
```diff
@@ -0,0 +1,282 @@
...
+                {
+                    // No Kinect, look for new one.
+                    if (Device.GetInstalledCount() > 0)
+                    {
+                        OnConnect();
+                    }
+                }
+                else
+                {
+                    // TODO: Handle Kinect disconnect. Device.GetInstalledCount every frame killed perf!
```
thegoldenmule: ?

Assets/Source/IoT/AzureKinect/KinectContext.cs
```diff
@@ -0,0 +1,282 @@
...
+        private void OnConnect()
+        {
+            try
+            {
+                _kinect = Device.Open();
+                _isConnected = true;
+            }
+            catch (DllNotFoundException e)
+            {
+                Log.Error(this, $"Missing required dll(s)! ({e.Message})");
```
thegoldenmule: I like that we are handling this separately: no string interp, no `e.`

Assets/Source/IoT/AzureKinect/KinectContext.cs
```diff
@@ -0,0 +1,282 @@
...
+                return;
+            }
+            catch (Exception e)
+            {
+                Log.Error(this, $"Unknown error (Is the device closed?): {e}");
```
thegoldenmule: no string interp, no `e.`

Assets/Source/IoT/AzureKinect/KinectContext.cs
```diff
@@ -0,0 +1,282 @@
...
+                return;
+            }
+            
+            Log.Info(this, $"Azure Kinect found: {_kinect.SerialNum}");
```
thegoldenmule: No string interp.

Assets/Source/IoT/AzureKinect/KinectContext.cs
```diff
@@ -0,0 +1,282 @@
...
+            if (_kinectCamera != null)
+            {
+                _cameraCulling = _mainCamera.cullingMask;
+                _mainCamera.cullingMask = 0;
+                
+                mode = CaptureProcessing.ProcessingMode.Combined;
+//                _kinectCamera.Setup(2560, 1440); // 16:9
+                _kinectCamera.Setup(3840, 2160); // 16:9
+//                _kinectCamera.Setup(2048, 1536); // 4:3
+//                _kinectCamera.Setup(4096, 3072); // 4:3
```
thegoldenmule: Should probably be configuration based. However! Maybe not part of the scene data as each TV may need different settings. Maybe this should be in a system menu.

Assets/Source/IoT/AzureKinect/KinectContext.cs
```diff
@@ -0,0 +1,282 @@
...
+                _kinectCamera.UpdateImages(images);
+            }
+
+            if (Time.realtimeSinceStartup - _lastMetricsSent > 5)
+            {
+                _metrics.Value($"{METRICS_PREFIX}.Temperature").Value(_captureProcessing.TemperatureFahrenheit);
+                _metrics.Value($"{METRICS_PREFIX}.ProcessingMs").Value(_captureProcessing.FrameTime);
+                _lastMetricsSent = Time.realtimeSinceStartup;
+            }
+//            Log.Info(this, _captureProcessing.FrameTime);
```
thegoldenmule: Delete.

Assets/Source/IoT/AzureKinect/KinectContext.cs
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        private void OnDisconnect()
+        {
+            Log.Info(this, "Closing Kinections");
```
thegoldenmule: I really appreciate this pun.

Assets/Source/IoT/AzureKinect/VideoWriter.cs
```diff
@@ -0,0 +1,202 @@
...
+                process.WaitForExit();
+
+                process.Dispose();
+            }
+            catch (Exception e)
+            {
+                Log.Error(this, process.StandardError.ReadToEnd());
+                process.Dispose();
+                
+                throw new Exception($"Error streaming to ffmpeg: {e}");
```
thegoldenmule: No string interp, no `e.`

Assets/Source/IoT/Kinect/BodyManagement.cs
```diff
@@ -0,0 +1,227 @@
+using System;
+using System.Collections.Generic;
+using CreateAR.Commons.Unity.Logging;
+using CreateAR.EnkluPlayer.IUX;
+using Enklu.Data;
+
+namespace CreateAR.EnkluPlayer.IoT.Kinect
+{
+    public class BodyManagement : IBodyManagement
```
thegoldenmule: Comments throughout.

Assets/Source/IoT/Kinect/BodyManagement.cs
```diff
@@ -0,0 +1,227 @@
...
+//            {
+//                Root = new ElementRef
+//                {
+//                    Id = root.Id
+//                },
+//                Elements = new [] { root },
+//            });
+//
+//            // track this body
+//            _bodies.Add(new BodyRecord(id, body));
```
thegoldenmule: Herp

Assets/Source/IoT/Kinect/IBodyManagement.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+
+namespace CreateAR.EnkluPlayer.IoT
+{
+    public interface IBodyManagement
```
thegoldenmule: Comments.

Assets/Source/IoT/Kinect/IBodyManagement.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        int MaxBodies { get; set; }
+        float CullDepth { get; set; }
+
+        void SetAssets(Dictionary<string, string> map);
```
thegoldenmule: Probably makes more sense as a property?

Assets/Source/IoT/Kinect/IKinectCommonContext.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+using CreateAR.EnkluPlayer.IUX;
+using CreateAR.EnkluPlayer.Scripting;
+
+namespace CreateAR.EnkluPlayer.IoT
+{
+    public interface IKinectCommonContext
```
thegoldenmule: Class comment. I am _assuming_ this is trying to provide some common functionality between Kinect types?

Assets/Source/IoT/Kinect/IKinectManager.cs
```diff
@@ -1,21 +1,41 @@
...
     /// <summary>
     /// Object that manages Kinect behavior.
     /// </summary>
     public interface IKinectManager
     {
+        /// <summary>
+        /// TODO: Refactor this. Avoids a circular dependency between KinectManager <-> ElementFactory
+        /// </summary>
+        /// <param name="elementFactory"></param>
+        void SetIElementFactory(IElementFactory elementFactory);
```
thegoldenmule: Kill with fire.

Assets/Source/IoT/Kinect/KinectContext.cs
```diff
@@ -1,5 +1,8 @@
-#if UNITY_STANDALONE_WIN
 
+
+using CreateAR.EnkluPlayer.IoT;
+using CreateAR.EnkluPlayer.IUX;
```
thegoldenmule: Seems like these `using` statements should be inside `#if`.

Assets/Source/IoT/Kinect/KinectContext.cs
```diff
@@ -134,19 +133,26 @@ public bool DebugDepth
...
             IBootstrapper bootstrapper,
+            IElementFactory elementFactory,
             IMessageRouter messageRouter,
             MainCamera mainCamera)
         {
             _bootstrapper = bootstrapper;
             _kinect = KinectSensor.GetDefault();
             _mainCamera = mainCamera.GetComponent<Camera>();
             
+            BodyManagement = new BodyManagement(elementFactory);
```
thegoldenmule: Not a cardinal sin, but it does make me question if `IBodyManagement` needs to exist.

Assets/Source/IoT/Kinect/KinectContext.cs
```diff
@@ -164,6 +170,11 @@ public bool DebugDepth
 
                 _kinect.Close();
             });
+            
+            // TODO: Stupid warnings as errors
+            OnBodyDetected?.Invoke(null);
+            OnBodyUpdate?.Invoke(null);
+            OnBodyLost?.Invoke(null);
```
thegoldenmule: I would pull this into another function or set of public functions, like `ForceDetected(BodyData data)`. This current implementation means that events will be dispatched.

Assets/Source/IoT/Kinect/KinectContext.cs
```diff
@@ -418,12 +428,14 @@ private void StartBody()
...
-            {
-                Body_OnLost(bodies[i]);
-            }
-            _body = null;
+//            var bodies = Bodies;
+//            for (var i = 0; i < bodies.Count; i++)
+//            {
+//                Body_OnLost(bodies[i]);
+//            }
+//            _body = null;
```
thegoldenmule: Delete or make work.

Assets/Source/IoT/Kinect/KinectJointTransform.cs
```diff
@@ -67,8 +67,10 @@ public void Update(Vec3 position, Quaternion rotation)
             _positions[_index] = position;
             _rotations[_index] = rotation;
 
-            Position = SmoothedPosition();
-            Rotation = SmoothedRotation();
+            Position = position;
+            Rotation = rotation.eulerAngles.ToVec();
+//            Position = SmoothedPosition();
+//            Rotation = SmoothedRotation();
```
thegoldenmule: Delete.

Assets/Source/IoT/Kinect/KinectManager.cs
```diff
@@ -1,9 +1,13 @@
-#if UNITY_STANDALONE_WIN
 
+using Object = UnityEngine.Object;
```
thegoldenmule: Should probably be inside `#if`.

Assets/Source/IoT/Kinect/KinectManager.cs
```diff
@@ -29,42 +39,82 @@ public class KinectManager : IKinectManager
...
             {
-                return new PassthroughKinectContext();
-            }
+                case KinectVersion.Any:
+                {
+                    return _ctx ?? (_ctx = new KinectContext(_bootstrapper, _elementFactory, _metrics, _messages, _mainCamera, profile, resources));
+                }
+                case KinectVersion.AzureKinect:
+                {
+                    return _ctx ?? (_ctx = new KinectContext(_bootstrapper, _elementFactory, _metrics, _messages, _mainCamera, profile, resources));
```
thegoldenmule: I would think that if `_ctx` is not null and is not an azure kinect, it should throw an exception or log a warning.

Assets/Source/IoT/Kinect/KinectManager.cs
```diff
@@ -29,42 +39,82 @@ public class KinectManager : IKinectManager
...
+                        return new PassthroughKinectContext();
+                    }
+
+                    if (!string.IsNullOrEmpty(kinectId) && kinect.UniqueKinectId != kinectId)
+                    {
+                        return new PassthroughKinectContext();
+                    }
 
-            return _ctx ?? (_ctx = new KinectContext(_bootstrapper, _messages, _mainCamera));
+                    return _ctx ?? (_ctx = new KinectContext(
```
thegoldenmule: Same here.

Assets/Source/IoT/Kinect/PassthroughKinectContext.cs
```diff
@@ -68,10 +71,12 @@ public bool Body(ulong id, out KinectSensorData data)
...
         public void ForceBodyDetected(ulong id)
         {
-            if (null != OnBodyDetected)
-            {
-                OnBodyDetected(id);
-            }
+            OnBodyDetected?.Invoke(null);
+        }
+
+        public void ForceBodyUpdate()
```
thegoldenmule: This is an example of what I was talking about to avoid warnings as errors.

Assets/Source/IoT/Kinect/PassthroughKinectManager.cs
```diff
@@ -14,13 +16,20 @@ public class PassthroughKinectManager : IKinectManager
         private readonly Dictionary<string, IKinectContext> _contexts = new Dictionary<string, IKinectContext>();
 
         /// <inheritdoc />
-        public IKinectContext Context()
+        public void SetIElementFactory(IElementFactory elementFactory)
```
thegoldenmule: For a passthrough it seems like this should not throw an exception. Though this function is going away so... :)

Assets/Source/Application/States/Play/HoloLensImageManager.cs
jsundra: It's probably better to treat this like `HoloLensVideoManager` - refactor out the uploading bits. `IVideoUploader` could maybe turn into somrthing like `IFileUploader` and have arguments/methods for images/videos

Assets/Source/IoT/AzureKinect/KinectContext.cs
jsundra: Configuration seems like it will bloat if we keep the practice of IoT devices in it.... Schema for a KinectWidget might not be too bad, but right now `KinectContext` can be created from scripting or Elements, so that could get tricky.... Yeah a system menu might be best. We can also through in the camera controls for exposure, brightness, etc in there.

Assets/Source/IoT/Kinect/IKinectCommonContext.cs
jsundra: Exactamundo!

Assets/Source/IoT/Kinect/KinectContext.cs
jsundra: I might end up removing `IBodyManagement`. So far it's kind of placeholder to see if v2 and Azure end up needing different interfaces. Hoping not though!

Assets/Source/Application/States/Play/IVideoCapture.cs
jsundra: 👍 Removed it from the interface. I'll go thru and add all the `#if UNITY_STANDALONE_WIN` for the Azure stuff, which probably should already be in there as things get more finalized

thegoldenmule:
@jsundra what's the status here? We need to get this in before we start major new features.

jsundra:
Planning to finish this up early next week!

Assets/Source/Application/States/Play/Standalone/WindowsVideoCapture.cs
jsundra: 👍 It was vague. Switched it up so `WindowsVideoCapture` owns creation and destruction - as opposed to just the optional destruction.

Assets/Source/Application/States/Play/Standalone/WindowsVideoCapture.cs
jsundra: Originally I had this because the recordings on windows don't capture any audio, and I thought it'd be good to know at the scripting level why a video that was setup with audio wouldn't have it. Changed it now to log a warning, and use the runtime's display size instead.

Down the road, I'd want to revisit this and the HoloLens implementation and combine them into one method with a config object passed.

Assets/Source/Application/States/Play/Standalone/WindowsVideoCaptureBootstrapper.cs
jsundra: Right now Kinects end up creating their own camera separate from the Main Camera and all of its configuration/monobehaviours, otherwise I'd have that as a MonoBleh and shoved it on the main camera.

I think that's a good practice to maintain for now of not reconfiguring the main camera at runtime and having to worry as much as reverting it and who owns it.

Assets/Source/Application/States/Play/Standalone/WindowsVideoCapture.cs
```diff
@@ -0,0 +1,237 @@
...
+        /// <summary>
+        /// Invoked every frame while videos are being recorded. Buffers captures from the GPU as needed.
+        /// </summary>
+        /// <param name="src">Source RenderTexture</param>
+        /// <param name="dest">Destination RenderTexture</param>
+        private void OnRenderImage(RenderTexture src, RenderTexture dest)
+        {
+            Graphics.Blit(src, dest);
+
+            if (Time.realtimeSinceStartup - _lastRenderTime >= _frametimeSeconds)
```
thegoldenmule: The way we're timing this is off, I believe. We need to have a target frame rate (say, 30 fps) and _accumulate_ dts. If we hit a high frametime right now, we could potentially not end up with enough frames for the time we're recording for, resulting in the entire video being sped up. If instead we accumulate dts, we can do a `while` loop here which will duplicate frames under processor intense conditions.

Assets/Source/Application/States/Play/Standalone/WindowsVideoCaptureBootstrapper.cs
thegoldenmule: I think this was originally referring to having a separate bootstrapper instead of just using a monobehavior.

Assets/Source/Application/States/Play/StandardVideoUploader.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+        
+        /// <inheritdoc />
+        public IAsyncToken<Void> UploadFile(string fullPath, string orgId, Dictionary<string, string> parameters)
+        {
+            var token = new AsyncToken<Void>();
+            var uploadTask = Upload(fullPath, orgId, parameters);
+            uploadTask.ContinueWith(run =>
+            {
+                if (run.IsFaulted)
```
thegoldenmule: Since a `Task` may call `ContinueWith` on another thread, I'm surprised we haven't seen threading issues here. There is an `ISynchronizer` that can resolve tokens on the main thread.

Assets/Source/Application/States/Play/VideoManager.cs
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        /// <summary>
+        /// Uploads a file from a given queue. Manages queue removal/addition as needed, as well as disk I/O.
+        /// </summary>
+        /// <param name="queue"></param>
+        /// <returns></returns>
+        private IAsyncToken<Void> UploadFromQueue(List<VideoPayload> queue)
+        {
+            var payload = queue[0];
```
thegoldenmule: I know you're checking for queue size above, but it seems to me like a better interface would be:

```
if (!UploadFromQueue(_uploadQueue, out token) && !UploadFromQueue(_retryQueue, out token)) {
  // failed
}
```

I bring this up because this function could very easily throw if the caller doesn't know to check queue size first. It's less of an issue because it's private, but it still makes me uneasy.

Assets/Source/IoT/AzureKinect/AzureKinectContext.cs
```diff
@@ -0,0 +1,389 @@
...
+            {
+                _mirror = value;
+                if (_kinectCamera != null)
+                {
+                    _kinectCamera.Mirror = _mirror;
+                }
+            }
+        }
+
+        public Vec2 MaxDepthEllipse
```
thegoldenmule: Comments here and below

Assets/Source/IoT/AzureKinect/BodyManagement.cs
```diff
@@ -0,0 +1,355 @@
...
+            _bodyMap.Clear();
+        }
+
+        /// <summary>
+        /// Invoked when a new body is detected.
+        /// </summary>
+        /// <param name="data"></param>
+        private void Body_OnDetected(BodyData data)
+        {
+            Log.Info(this, "Body detected: {0}", data.BodyId);
```
thegoldenmule: We should move these to debug or remove.

Assets/Source/Player/Scripting/Interface/Kinect/PostProcessingJs.cs
```diff
@@ -0,0 +1,169 @@
+﻿using Enklu.Data;
+using UnityEngine;
+using UnityEngine.Rendering.PostProcessing;
+
+namespace CreateAR.EnkluPlayer.Scripting
+{
+    public class PostProcessingJs
```
gdamron: xml docs

Assets/Source/Application/States/Play/Standalone/WindowsVideoCapture.cs
jsundra: Hmm.... I think accumulating dts would still run into a problem with a large frame though, unless it was running on its own thread and capturing frames differently? I'm actually not sure if the `RenderTexture`s Unity passes are always reused instances or if they ever rotate either, so keeping references and acting on them later might be problematic. Unless I'm misunderstanding what you're suggesting?

thegoldenmule: It won't fix the large frame issue, no. That is external to video recording-- that could happen if there's just too much to render or a CPU intensive script or something. You wouldn't need to save frame data, just do `_frames.Add(data)` with the same data adding up to the _real_ dt.

For example, say we want to capture a frame every 33 ms but a frame comes along that's 90. With the current approach you miss 2 entire frames, meaning that the resulting video will be sped up (as it's putting frames in every 33 ms). Instead, you would add 3 of the same frame, i.e. you accumulate the same frame up to the real dt.

jsundra: Ahhhh gotcha. Makes sense! I thought you were meaning to run the framegrabber in a coroutine with its own `while(true)` or equivalent.

Yeah, I can see where the video frames will be incorrectly spaced. Good catch!

Assets/Source/Application/States/Play/VideoManager.cs
jsundra: Hmm... Yeah I don't like the chance to throw. I'm not sure how I feel about that interface though, at first glance I would think that `UploadFromQueue` returning true would mark a success, and not a start of something that could fail. In my head I'm comparing it to `Dictionary.TryGetValue` which makes sense because it's all or nothing - albeit synchronous.

How would you feel about the returned token being failed when it normally would throw an exception instead?

Assets/Source/Application/States/Play/StandardVideoUploader.cs
```diff
@@ -0,0 +1,145 @@
...
+                Log.Info(this, "Uploading {0} to {1}", filepath, url);
+
+                var httpClient = new HttpClient();
+                foreach (var kvp in _http.Services.GetHeaders("trellis"))
+                {
+                    httpClient.DefaultRequestHeaders.Add(kvp.Key, kvp.Value);
+                }
+                httpClient.DefaultRequestHeaders.Add("Accept", "application/json");
+
+                var form = new MultipartFormDataContent("Boundary7MA4YWxkTrZu0gW");
```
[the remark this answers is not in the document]
jsundra: 👍 I can definitely pull that up to a const, where it should be. That boundary specifically though, isn't meant to communicate specifically with any of our other services right? It's just meant to have enough entropy to ensure it is unique - at least that's my understanding of boundaries :)

Assets/Source/IoT/AzureKinect/CaptureProcessor.cs
```diff
@@ -0,0 +1,267 @@
...
+            Calibration = null;
+            _kinect.StopCameras();
+        }
+
+        /// <summary>
+        /// Capture processing. Happens on its own independent thread.
+        /// </summary>
+        private void ProcessingThread()
+        {
+            Log.Info(this, "Spinning up processing thread.");
```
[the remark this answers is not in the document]
jsundra: `GetCapture` definitely is blocking. You're right that it isn't actually spinning up the thread here - but I lean towards wanting the log after it is spun up and begins processing, rather than when we're calling `ThreadHelper` to start it since there's other logs before then. I think I'll change the phrasing to be something like `Processing thread spun up.` - something past tense.

Assets/Source/IoT/AzureKinect/VideoWriter.cs
```diff
@@ -0,0 +1,181 @@
...
+            var outputArgs = string.Format(OUTPUT_ARGS, filename);
+
+            if (!string.IsNullOrEmpty(config.AudioFile))
+            {
+                // TODO: Add -t to outputArgs
+                inputArgs += $" -i {config.AudioFile} -map 0:v:0 -map 1:a:0";
+            }
+            
+            Log.Info(this, @"Creating ffmpeg process. 
+Filename={0}
```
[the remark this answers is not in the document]
jsundra: Yeah it isn't the prettiest, but it keeps the log clean of tabs/spaces the verbatim string would add. Having something like `$"Creating ffmpeg process.\nFilename={filename}\nInput={inputArgs}\nOutput={outputArgs}" definitely feels way too ugly. Do you have a cleaner solution in mind?

Assets/Source/Player/Scripting/Interface/Kinect/PostProcessingJs.cs
jsundra: Dang, thought I got 'em all :(

Assets/Source/Application/States/Play/VideoManager.cs
thegoldenmule: It seems clear enough to me, but that's what these reviews are for: to make sure it's clear to everyone. Returning a failed token also seems clear to me, so go nuts.

Assets/Source/Application/States/Play/StandardVideoUploader.cs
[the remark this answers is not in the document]
thegoldenmule: Boundaries are usually generated, but in this case I got lazy and just used a boundary from a successful HTTP request I made to make sure it would work. That is, it was easier to use a const that I knew worked than prove that whatever generative function I made worked.

Assets/Source/IoT/AzureKinect/VideoWriter.cs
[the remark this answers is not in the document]
gdamron: I won't block on this since it's partly a matter of preference. I'd personally rather have two calls to `Log.Info` if the desire is a multiple line log entry. It makes for less weird indentation in the logs themselves.

PR #326, opened 2019-10-14
Hotfix/ms store

thegoldenmule:
- WSA store verification bug prevents store builds from using ChakraCore.dll. `#if`'d out the references.
- Fixed Kinect import settings so HL builds do not include it.
- Fixed issue where load timeout could occur with scenes that had not previously been loaded, in which case load from disk failed. Now, disk loads _can_ resolve early but cannot kill network load.
- New certificate.


thegoldenmule:
The Windows certification fails because ChakraCore.dll references some things that are apparently off limits for store builds.

PR #329, opened 2019-11-12
[T#95] Support for overriding settings in individual builds

gdamron:
This branch adds a new source for loading configuration at application start time. This file can be placed at `<unity-persistent-data-dir>/Config/override.prefs`. There are a few things worth noting:

- This override is applied in a manner identical to `ApplicationConfig.Override.json`, so some fields can get reset to defaults if they were overridden at build time (i.e.: the default value for `Metrics.Enabled` is `true`, so setting it to `false` in `ApplicationConfig.Override,json` and then setting `Metrics.ApplicationKey` in `override.prefs` will result in a final `true` value for `Metrics.Enabled`
- It is loaded last, so it always wins
- A use of `JsonSerializer` was converted to Newtonsoft for consistency
- This PR includes a conversion to Unix-style line endings (as per the style guide). I recommend adding the `&w=1` query param to this page to filter out some of the noise,


gdamron:
docs updated!

PR #330, opened 2019-11-21
Update playerVersion

barrygear:
Update playerVersion
chmod +x for gradlew


thegoldenmule:
In the future we need to be sure to:

- Make sure the branch off of master is prefixed by `hotfix/`.
- We use semver in general, though the Windows Store requires the build number tacked on the end. Because of this, we leave the build number as zero and always increment either minor or bugfix.

PR #332, opened 2019-11-21
Multiplayer Notifications

thegoldenmule:
Integrated notification API. This allows users to broadcast arbitrary string payloads.


thegoldenmule:
See https://github.com/enklu/enkluplayer-jsapi-docs/pull/16

Assets/Source/Application/States/Play/Multiplayer/WebMultiplayerController.cs
```diff
@@ -163,6 +172,14 @@ public void Forfeit(string elementId)
             // 
         }
 
+        /// <summary>
+        /// Allows directly sending notification.
+        /// </summary>
+        public void Notify(string type, string payload)
```
jsundra: Is this unused?

Assets/Source/Application/States/Play/Multiplayer/WebMultiplayerController.cs
```diff
@@ -151,6 +154,12 @@ public void UnSync(string elementId, ElementSchemaProp prop)
             // 
         }
 
+        /// <inheritdoc />
+        public void Broadcast(string type, string payload)
+        {
+
```
jsundra: I know mycelium doesn't connect in the web editor, but it'd be really helpful to have `Broadcast` trigger notifications still for the editor client. Similar to how a lot of the `autoToggle` functionality has the offline fallbacks that are usable in the editor.

thegoldenmule: Broadcasts are not sent to the sender, so this reflects the current behavior.

Assets/Source/Application/States/Play/Multiplayer/WebMultiplayerController.cs
thegoldenmule: Yes. There is a warning about `OnNotification` being unused. The two easiest fixes are to either add a `pragma` to disable the warning in this specific case or to add a public usage of it. I settled on adding a public usage as I think that's the cleaner of the two options.

gdamron: I'm sort of on the fence with this method.

Would make sense to add this to `IMultiplayerController` (and the other implementation) and expose it via JavaScript? That would address @jsundra 's concern about testing notifications in the editor. I can also imagine scenarios where a client would want to send itself a message through the same channel as multiplayer stuff.

On the other hand, this sort of self-notification strays a bit from the class's multiplayer responsibilities. In that case, maybe it's better to simply remove it until there is a compelling use case for it to be here.

If if stays, I think xml comment should clarify that it sends a message to the current client. My first reading was that it was a way to notify a specific, _different_ client

thegoldenmule: I believe we're talking about two different things. Testing notifications in the web editor is unrelated to this particular method. This method does not define any new functionality and is not intended to be used to broadcast notifications in the editor. Let me split into two threads:

1. **Should broadcast call the local client?** I don't believe so. The multiplayer model is based on chat rooms. In a chat room, a broadcast is outgoing only-- the server does not send you a message when you @channel in slack. Many chat implementations (I'm thinking of XMPP) have a broadcast and a _whisper_. The whisper is essentially a direct message, and the broadcast is to the room. The backend supports whispers (see https://github.com/enklu/enkluplayer/pull/332/commits/c1dd259b693c1acf5f429b978a69b1b7dce93c8e) but I have elided them for now. This is not only convention, but reduces unnecessary network traffic.

2. **Should we introduce a public function to get around a compile warning?** I vote yes in this case precisely because the public function is on the implementation rather than the interface. This means that it is invisible to all users except for integration or unit tests that may be testing this specific implementation.

Assets/Source/Application/States/Play/Multiplayer/WebMultiplayerController.cs
jsundra: Happy with how it is then :)

Assets/Source/Application/States/Play/Multiplayer/WebMultiplayerController.cs
gdamron: Agree with #1, that seems pretty clear to me.

#2, though, I'm not so sure. Since it's unused, not a part of the interface, and it's not driven by a particular need, I'd vote to remove it -- unless you're arguing there is a specific reason for it being there.

thegoldenmule: This brings us back to my initial response. Since warnings give us compile errors, we can either add a pragma around the unused event `OnNotification` or we can add a usage for it. I opted for the latter.

gdamron: my objection is more about philosophy than compilation. code that is available for use but not _actually_ used is clutter -- regardless of whether our tools object

gdamron: but I'll stop there since it's such a small snippet of code :)

gdamron: adding notes from an offline conversation with @thegoldenmule here:

If there is a good way to add the `Notify` method to the test suite, that's what we should do. Otherwise, the current implementation is the cleanest option

PR #344, opened 2019-12-13
JS login api

gdamron:
T#1355 Create JS api to log out a user
T#1354 Create JS API to show login screen

If an experience developer calls `system.logout()`, the user will be logged out immediately. The current behavior is that they are pushed back through the login flow.

If an experience developer calls `system.login()` they will be pushed through the log in flow they are not yet logged in. If the user is already logged in, nothing happens. There is an optional `force` parameter as well. if true (`system.login(true)`), the user will be forcefully logged out and pushed through the login flow.


Assets/Source/Player/Scripting/Interface/App/SystemJsApi.cs
```diff
@@ -200,6 +212,40 @@ public void restart()
...
+        }
+
+        /// <summary>
+        /// Logs out a user.
+        /// </summary>
+        public void logout()
+        {
+            Log.Info(this, "Logging out user");
+            _messages.Publish(MessageTypes.SIGNOUT);
+        }
```
thegoldenmule: This is looking great. I do think we should alter the interface slightly: not a huge fan of the `bool` parameter. If I read `login(true)`, I would not think "yah this is going to log me out." How about:

```
login()
logout()
isLoggedIn
```

A cleaner interface that can handle edge cases well (`login` and `logout` can silently do nothing).

Assets/Source/Player/Scripting/Interface/App/SystemJsApi.cs
```diff
@@ -200,6 +212,40 @@ public void restart()
...
+            _messages.Publish(MessageTypes.LOGIN);
+        }
+
+        /// <summary>
+        /// Logs out a user.
+        /// </summary>
+        public void logout()
+        {
+            Log.Info(this, "Logging out user");
+            _messages.Publish(MessageTypes.SIGNOUT);
```
thegoldenmule: I think this should also check if you are logged in and do nothing if you are not.

Assets/Source/Player/Scripting/Interface/App/SystemJsApi.cs
gdamron: agreed 👍

Assets/Source/Player/Scripting/Interface/App/SystemJsApi.cs
gdamron: 👍

PR #347, opened 2019-12-16
Bugfix/scripts in streaming assets

gdamron:
When running `Tools > Experiences > Save to Streaming Assets`, include script content in `Streaming Assets/Scripts`


Assets/Editor/Tools/Experiences/AppManifestHooks.cs
```diff
@@ -230,7 +230,16 @@ private static IAsyncToken<Void> GetScriptsForApp(string appId)
             _httpCache.Request(_behavior,
                     "streaming://" + appId + "/scripts",
                     () => _api.PublishedApps.GetPublishedAppScripts(appId))
-                    .OnSuccess((res) => token.Succeed(Void.Instance));
+                    .OnSuccess((res) =>
+                    {
+                        var scripts = res.Body;
+                        foreach (var script in scripts)
```
thegoldenmule: I think `foreach` is fine in editor code btw, but it often makes its way into the play code.

PR #348, opened 2019-12-16
Load stored credentials at launch

gdamron:
Load stored authentication during initialization.

Note that the stored app id is excluded at this stage.


Assets/Source/Application/ApplicationConfigExtensions.cs
```diff
@@ -23,5 +19,35 @@ static public bool IsLoggedIn(this CredentialsData @this)
...
+            http.Services.Urls.Formatter("trellis").Replacements["userId"] = @this.UserId;
+            http.Services.AddHeader("trellis", "Authorization", string.Format("Bearer {0}", @this.Token));
+        }
+
+        /// <summary>
+        /// Applies credentials to the Application Config.
+        /// </summary>
+        /// <param name="this"></param>
+        /// <param name="config"></param>
+        static public void Apply(this CredentialsData @this, ApplicationConfig config)
```
thegoldenmule: Should be `public static`. Is Resharper not asking you to change this?

gdamron: Thanks for catching that. Looks like my ReSharper license is out of date.

PR #350, opened 2019-12-16
Allow asset bundles to be preloaded

gdamron:
Adds `Preload` option to `ApplicationConfig.PlayConfig`. If true, assets bundles will not be loaded lazily. Instead, they will all be loaded before entering play mode.


Assets/Source/Application/States/Play/App/AppController.cs
```diff
@@ -230,6 +257,47 @@ public void Edit()
...
+        /// <returns></returns>
+        private IAsyncToken<Void> PreloadAssets()
+        {
+            var retToken = new AsyncToken<Void>();
+
+            var assets = new HashSet<string>();
+            foreach (var sceneId in _loader.Scenes)
+            {
+                var description = _loader.Scene(sceneId);
+                foreach (var elementData in description.Elements)
```
thegoldenmule: These should all be `for`s to avoid allocation.

Assets/Source/Application/States/Play/App/AppController.cs
```diff
@@ -230,6 +257,47 @@ public void Edit()
...
+                {
+                    assets = AppDataLoader.ExtractAssetIds(elementData, assets);
```
thegoldenmule: Kind of confusing here, is this the best way we can do this? We're passing in a hashset AND returning one.

Assets/Source/Application/States/Play/App/AppController.cs
```diff
@@ -230,6 +257,47 @@ public void Edit()
...
+                assetTokens.Add(token);
+            }
+
+            Async.All(assetTokens.ToArray())
+                .OnSuccess(loadedAssets =>
+                {
+                    _isLoaded = true;
+                    retToken.Succeed(Void.Instance);
+                })
+                .OnFailure(err =>
```
thegoldenmule: Style recommendation (take or leave):

`.OnFailure(retToken.Fail);`

Assets/Source/Application/States/Play/App/AppDataLoader.cs
```diff
@@ -165,6 +165,35 @@ public static ScriptData ToScriptData(Trellis.Messages.GetPublishedAppScripts.Bo
             };
         }
 
+        /// <summary>
+        /// Recursively creates a set of all asset ids for an element and its children.
+        /// </summary>
+        /// <param name="element"></param>
+        /// <param name="assetIds"></param>
+        /// <returns></returns>
+        public static HashSet<string> ExtractAssetIds(ElementData element, HashSet<string> assetIds = null)
```
thegoldenmule: I would recommend simplifying this function a lot: returning null and requiring the caller to pass in the hashset. There's no real reason to allocate inside this function and it makes the function more confusing from the outside.

Assets/Source/Application/States/Play/App/AppDataLoader.cs
```diff
@@ -165,6 +165,35 @@ public static ScriptData ToScriptData(Trellis.Messages.GetPublishedAppScripts.Bo
...
+
+            var assetKey = "assetSrc";
+            var strings = element.Schema.Strings;
+            if (strings.ContainsKey(assetKey) && !string.IsNullOrEmpty(strings[assetKey]))
+            {
+                var asset = strings[assetKey];
+                assetIds.Add(asset);
+            }
+
+            foreach (var child in element.Children)
```
thegoldenmule: `for`

Assets/Source/Player/Asset/LocalAssetLoader.cs
```diff
@@ -15,7 +17,22 @@ public class LocalAssetLoader : ILocalAssetLoader
         /// <inheritdoc cref="ILocalAssetLoader"/>
         public IEnumerator Load(string bundle)
         {
+            Bundle = null;
```
thegoldenmule: This looks like a leak. Don't we need to do `Bundle.Destroy()` (or Unload-- I don't recall).

Assets/Source/Player/Asset/LocalAssetLoader.cs
```diff
@@ -15,7 +17,22 @@ public class LocalAssetLoader : ILocalAssetLoader
...
+
+            var loadedBundles = AssetBundle.GetAllLoadedAssetBundles();
+            var loaded = loadedBundles.FirstOrDefault(a => a.name == bundle);
```
thegoldenmule: LINQ allocates like crazy-- we should pull that out and just write out the `for` loop.

Assets/Source/Application/States/Play/App/AppDataLoader.cs
gdamron: Sure, will make it a void function, remove the unnecessary parts, and clarify in the docs that it mutates the `assetIds` set

Assets/Source/Application/States/Play/App/AppDataLoader.cs
gdamron: 👍

Assets/Source/Application/States/Play/App/AppController.cs
gdamron: 👍

Assets/Source/Application/States/Play/App/AppController.cs
gdamron: see other comment

Assets/Source/Application/States/Play/App/AppController.cs
gdamron: 👍

Assets/Source/Player/Asset/LocalAssetLoader.cs
gdamron: 👍 yes, thanks for catching that!

Assets/Source/Player/Asset/LocalAssetLoader.cs
gdamron: 👍

Assets/Source/Application/States/Play/App/AppController.cs
gdamron: although IMHO -- that style, while less verbose, makes debugging more difficult since you can't set a breakpoint on it

Assets/Source/Player/Asset/LocalAssetLoader.cs
gdamron: Although... we don't want to unload the assets here. this is a reference to a bundle that the `AssetBundleLoader` ultimately manages. There _is_ a danger of a leak here, though. One possibility would be to replace the `Bundle` object with a `GetBundle` method. Then, if `GetBundle` has not been called before the next call to `Load`, the current bundle can be unloaded.

Assets/Source/Application/States/Play/App/AppController.cs
thegoldenmule: True

Assets/Source/Application/States/Play/App/AppDataLoader.cs
thegoldenmule: :+1:

Assets/Source/Player/Asset/LocalAssetLoader.cs
thegoldenmule: :+1: I like it, good catch on who owns what!

PR #351, opened 2019-12-16
Try to load scripts from streaming assets

gdamron:
If a script exists in the Streaming Assets folder, attempt to load that copy first.


Assets/Source/Player/App/StandardScriptCache.cs
```diff
@@ -49,9 +57,17 @@ public virtual void Save(string id, string version, string value)
         /// <inheritdoc />
         public virtual IAsyncToken<string> Load(string id, string version)
         {
-            return _files
-                .Get<string>(Uri(id, version))
-                .Map(file => file.Data);
+            var token = new AsyncToken<string>();
+            _files
```
thegoldenmule: Since this class is already in the weeds of checking if files exist in `Contains`, seems like we could shortcut this with a quick `if (_files.Exists())` rather than the over head of multiple async operations.

gdamron: 👍

PR #354, opened 2019-12-18
ensure player subscribes to updates when switching experiences

gdamron:
forces web sockets to tear themselves and reconnect when switching apps


Assets/Source/Application/Connection/UwpConnection.cs
```diff
@@ -75,7 +75,22 @@ public IAsyncToken<Void> Connect(EnvironmentData environment)
...
+
+                    IsConnected = false;
+
+                    _connectToken.Fail(new Exception("Socket closed."));
+                    _connectToken = null;
+                }
+                catch (Exception exception)
+                {
+                    LogVerbose("Could not dispose socket: {0}.", exception);
+                }
```
thegoldenmule: I think the `try/catch` should be around only `_socket.Dispose()`. The way it is here, if a caller adds an `OnFailure` handler that throws an exception, it will be caught here and assumed to be a dispose issue.

Assets/Source/Application/Connection/WebSocketSharpConnection.cs
```diff
@@ -174,6 +174,14 @@ private IAsyncToken<Void> ConnectSocket(string wsUrl)
 
             return token.Token();
         }
+
+        private void ReleaseSocket()
+        {
+            _connectToken.Fail(new Exception("Socket closed."));
+            _connectToken = null;
+
+            IsConnected = false;
```
thegoldenmule: This function doesn't release a socket. Should it be renamed?

Assets/Source/Application/Connection/UwpConnection.cs
gdamron: 👍

Assets/Source/Application/Connection/WebSocketSharpConnection.cs
gdamron: will change. something like `ClearConnectToken`?

thegoldenmule: Sounds good to me.

PR #355, opened 2019-12-18
Use existing login flow as much as possible and limit asset preloading

gdamron:
- Restores the loading of cached credentials to be the responsibility of `LoginApplicationState`
- `HmdPublicPublicAppStateFlow` will now check for stored credentials before changing state
- The preload option in `ApplicationConfig.PlayConfig` is limited to the app override rather than being an application wide setting

While the intention is to ensure F#1290 (Doesn't require WiFi), this also fixes B#1539 (Live Updates not working)


Assets/Source/Application/Flows/HmdPublicAppStateFlow.cs
```diff
@@ -60,10 +67,21 @@ public void Start(IApplicationStateManager states)
                 MessageTypes.DEVICE_REGISTRATION_COMPLETE,
                 MessageTypes.SIGNOUT);
 
-            // Start in play mode. the user may not be logged in, and we don't
-            // know if they have permission to edit
+            // Start in play mode.
             _config.Play.Edit = false;
-            _states.ChangeState<LoadDefaultAppApplicationState>();
+
+            if (_files.Exists(LoginApplicationState.LOGIN_URI))
```
jsundra: Down the road, it'd be great to have a helper for login/user/session management - so that non-login specific flows don't need to know about login files. But for now, this works :)

gdamron: agreed. i almost went down that road for this but decided it deserved more thought. i'll add a TODO

PR #359, opened 2019-12-20
Added speech controller on platforms that don't support Azure speech

gdamron:
Opening on behalf of @thegoldenmule 

Looks fine to me. @jsundra any comments?


thegoldenmule:
Nice work, me.

PR #360, opened 2019-12-28
B#1111 User assets sometimes don't load

gdamron:
- Check if an asset is already in the manifest before adding to avoid an exception
- All '-' characters in schema queries in order to search for guids
- Signal that elements using an asset should reload after adding an asset since there is no guarantee before that the asset was added before the scene was updated. If it was already updated, the asset could not have loaded.


gdamron:
also to master...

PR #361, opened 2019-12-28
B#1591 Poor performance in web player when light gizmos enabled

gdamron:


gdamron:
oops, that was to master

PR #363, opened 2019-12-28
B#1111 User assets sometimes don't load

gdamron:
- Check if an asset is already in the manifest before adding to avoid an exception
- Allow '-' characters in schema queries in order to search for guids
- Signal that elements using an asset should reload after adding an asset since there is no guarantee before that the asset was added before the scene was updated. If it was already updated, the asset could not have loaded.


Assets/Source/Player/Asset/AssetManifest.cs
```diff
@@ -209,6 +209,16 @@ public void Destroy()
             _guidToRecord.Clear();
         }
 
+        /// <summary>
+        /// Returns true if an asset exists in the manifest.
+        /// </summary>
+        /// <param name="asset"></param>
+        /// <returns></returns>
+        public bool Contains(AssetData asset)
```
thegoldenmule: Please add appropriate unit test coverage for this, as `AssetManifest` is a central system.

Assets/Source/Player/IUX/Element/ElementQuery.cs
```diff
@@ -22,7 +22,7 @@ internal class ElementQuery
         /// <summary>
         /// Regex for property matches.
         /// </summary>
-        private static readonly Regex PROPERTY_QUERY = new Regex(@"^\(@([\w-_\.]+)([<>=!]=?)(\w+)\)$");
+        private static readonly Regex PROPERTY_QUERY = new Regex(@"^\(@([\w-_\.]+)([<>=!]=?)([\w-]+)\)$");
```
thegoldenmule: Add a unit test that tests guid lookup.

Assets/Source/Player/Asset/AssetManifest.cs
gdamron: 👍

Assets/Source/Player/IUX/Element/ElementQuery.cs
gdamron: 👍

gdamron:
yes, all unit tests still pass with the changes. it's worth pointing out that our current tests use values like "guid" or single letters for guids, which is why they never illustrated that searching for an _actual_ guid fails.

gdamron:
tests updated

PR #365, opened 2020-01-09
Player Restarts

thegoldenmule:
This PR fixes the flow for essentially restarting Enkluplayer while Unity is still running. It uninitializes the `Application` instance on restart events, then reinitializes. This is needed especially in the web player when we load between scenes, but will also be useful as we move toward the plugin.


Assets/Source/Application/States/Play/App/HttpRequestCacher.cs
```diff
@@ -51,19 +51,13 @@ public HttpRequestCacher(IFileManager files)
...
 
             // first attempt to load from disk with override key.
-            _files.Get<T>(overrideKey)
-                .OnSuccess(file =>
-                {
-                    token.Succeed(file.Data);
-                })
-                .OnFailure(err1 =>
-                {
-                    Log.Info(this, "First attempt at loading from disk failed.");
```
thegoldenmule: This log doesn't give any good info, so I removed it.

Assets/Source/Application/Bridge/WebBridge.cs
```diff
@@ -52,17 +55,24 @@ public void Initialize(BridgeMessageHandler handler)
...
+            if (null == _logTarget)
+            {
+                _logTarget = new JsBridgeLogTarget(this);
+            }
+
 #if !UNITY_EDITOR && UNITY_WEBGL
             ready();
 
-            Log.AddLogTarget(new JsBridgeLogTarget(this));
+            Log.AddLogTarget(_logTarget);
```
jsundra: `BroadcastReady` vs `Uninitialize` doesn't seem like there's a lot of parity in the naming. Rather, it almost sounds like the former may be called multiple times due to its name, which would add the log target multiple times. Maybe I'm just being picky though?

thegoldenmule: The flow is `Initialize` -> `BroadcastReady` -> `Uninitialize` -> Repeat.

You're right that the bridge doesn't enforce the flow, perhaps it could. It would be nice to rethink the interface and make it a tad more clear, but I think it's fairly low priority right now, and certainly not part of this PR.

PR #368, opened 2020-01-13
Add support for sending metrics to our own backend

gdamron:
- Not as big as it looks. Most of this is updating the stargazer API
- Note the "breaking" change to `ApplicationConfig`


Assets/Source/Application/ApplicationConfig.cs
```diff
@@ -858,62 +858,118 @@ public void Override(ConductorConfig config)
         }
     }
 
-    /// <summary>
-    /// Configuration for metrics.
-    /// </summary>
-    public class MetricsConfig
+    public class MetricsAppConfig
```
thegoldenmule: Comments.

Assets/Source/Application/ApplicationConfig.cs
```diff
@@ -858,62 +858,118 @@ public void Override(ConductorConfig config)
...
 
         /// <summary>
-        /// Overrides settings.
+        /// Applies the settings from a config to this config. All objects in the
+        /// Targets array should have a non-empty value for Targets.
         /// </summary>
-        /// <param name="config">The config.</param>
-        public void Override(MetricsConfig config)
+        /// <param name="overrideConfig">The config to override with.</param>
+        public void Override(MetricsAppConfig overrideConfig)
```
thegoldenmule: To think about for later: this is getting a little out of control. It might be time to start poking around using an existing, battle-tested C# configuration system.

Assets/Source/Application/Services/MetricsUpdateService.cs
```diff
@@ -89,22 +95,22 @@ public override void Stop()
         /// Updates the enabled metrics datapoints per the configuration's interval.
         /// </summary>
         /// <returns></returns>
-        private IEnumerator UpdateData()
+        private IEnumerator UpdateData(MetricsDataConfig config)
         {
             var id = _currentId;
             
             while (id == _currentId)
```
thegoldenmule: Since `UpdateData` is now being called multiple times, it breaks this approach of checking `_currentId`. With this implementation, only the last coroutine started will continue running.

Assets/Source/Main.cs
```diff
@@ -146,6 +152,22 @@ private void Start()
...
+                    try
+                    {
+                        target.Setup(configTarget.ApplicationKey);
+                        metrics.AddTarget(target, MetricTargetType.User);
+                    }
+                    catch (Exception ex)
+                    {
+                        Log.Warning(this, "Could not set up EnkluMetricsTarget: {0}", ex);
+                    }
+                }
```
thegoldenmule: Probably about time to move metric target configuration into a separate object or at least a separate method.

Assets/Source/Stargazer/Verify.Response.cs
```diff
@@ -0,0 +1,34 @@
+// To parse this JSON data, add NuGet 'Newtonsoft.Json' then do:
```
thegoldenmule: More future stuff to think about: pulling all this crud into a dll.

Assets/Source/Util/Analytics/MetricsService.cs
```diff
@@ -27,14 +33,49 @@ public class MetricsService : IMetricsService
...
+                _ =>
+                {
+                    for (var i = 0; i < _timers.Count; i++)
+                    {
+                        var timer = _timers[i];
+                        timer.Pause();
+                    }
+                });
+
+            messages.Subscribe(MessageTypes.APPLICATION_RESUME,
```
thegoldenmule: HAAAAAAAAAAAAAAAH

thegoldenmule: Actually-- from what I can tell, suspend doesn't fire until after the app resumes. It's a bit... odd.

Assets/Source/Util/Analytics/MetricsService.cs
```diff
@@ -81,7 +152,22 @@ public ValueMetric Value(string key)
...
+                }
+                default:
+                {
+                    target = _deviceTarget;
+                    break;
+                }
+            }
+
+            var newValue = new ValueMetric(target, key);
             _values.Add(newValue);
```
thegoldenmule: Having MetricsService implement everything with switches seems pretty brittle. I know there are only 2 enums but a general solution seems like it would be simpler.

Assets/Source/Application/ApplicationConfig.cs
gdamron: 👍

Assets/Source/Application/ApplicationConfig.cs
gdamron: yep, absolutely agree

Assets/Source/Application/Services/MetricsUpdateService.cs
gdamron: ah, thanks, good catch

Assets/Source/Main.cs
gdamron: 👍

Assets/Source/Stargazer/Verify.Response.cs
gdamron: yes!

Assets/Source/Util/Analytics/MetricsService.cs
gdamron: well, the same method gets called when suspending and resuming. it is definitely a bit odd, though

Assets/Source/Util/Analytics/MetricsService.cs
gdamron: yes, and i consciously punted on that for now to give it a little more thought. at the very least, i can add a `TODO`. i can also start down that road now, if you'd like, and am happy to accept any thoughts

thegoldenmule: A TODO is fine.

Assets/Source/Application/Services/MetricsUpdateService.cs
gdamron: Actually, I don't think that is the case. Since the id is only incremented on `Stop`, all coroutines should run. It's one shared id for the service rather than individual targets. Or am I missing something?

thegoldenmule: Ah, I think you're right!

Assets/Source/Main.cs
```diff
@@ -120,15 +120,6 @@ private void Start()
 
             MetricsTargetHelper.AddTargets(configTargets, metrics, _Binder);
             
-            // handle restarts
-            _Binder.GetInstance<IMessageRouter>().Subscribe(
```
gdamron: @thegoldenmule  please confirm this was removed by you in a previous commit

thegoldenmule: Yes, I moved that somewhere else.

PR #369, opened 2020-01-13
Feature/misc fixes

jsundra:
Few things:
- Fixed a bug with HL2 hand tracking not always returning the correct hand's data.
- Added `q.euler` overload for parity with `q.eul`.
- Fixed Kinect v2 rotation.
- Surfaced Kinect Id to scripting.


Assets/Source/Player/Scripting/Interface/HandJsInterface.cs
```diff
@@ -143,32 +147,40 @@ public void off(string @event, string hand, IJsCallback cb)
...
-            return getRotation((Handedness) hand, (TrackedHandJoint) Enum.Parse(typeof(TrackedHandJoint), joint));
-        }
-
-        private Vec3 getPosition(Handedness hand, TrackedHandJoint handJoint)
-        {
-            return _latestData[handJoint].Position;
+            try
+            {
+                var trackedHandJoint = (TrackedHandJoint) Enum.Parse(typeof(TrackedHandJoint), joint);
+                return _latestData[handedness - 1][trackedHandJoint].Position;
```
thegoldenmule: This will throw an unintelligible error to the scripter. We should explicitly check bounds and log out specific errors like, "handedness has to be 1 or 2" and "that is not a valid joint.".

Assets/Source/Player/Scripting/Interface/HandJsInterface.cs
```diff
@@ -143,32 +147,40 @@ public void off(string @event, string hand, IJsCallback cb)
...
+                Log.Warning(this, "Error getting data for hand: {0} {1} {2}", handedness, joint, exception);
+            }
+            return Quat.Identity;
         }
 
         /// <inheritdoc />
         public void HandDetected(Handedness handedness, Dictionary<TrackedHandJoint, JointTransform> data)
         {
-            _latestData = data;
+            _latestData[(int) handedness - 1] = data;
```
gdamron: I believe the value for unspecified handedness is 0, so you're potentially using -1 as an index

jsundra: Yeah.... the underlying API won't return a `None` hand.... but that doesn't really make this code clean. I'll fix it up!

PR #372, opened 2020-01-14
TP#1760 - Fix issue where sending across bridge threw error if sent before player was loaded

thegoldenmule:
Consume send queue after loaded.


thegoldenmule:
See https://github.com/enklu/editron-5000/pull/362

PR #374, opened 2020-01-22
Undefined ContentWidget JsAPI bugfix

jsundra:
Found the fix while reproing :) In the past we had a specific event when assets were finished assembling, which changed recently to be when assets + scripts finished. This would cause all APIs reliant on `this.*` to be undefined when the player first loads.


Assets/Source/Player/IUX/Content/ContentWidget.cs
```diff
@@ -81,10 +81,15 @@ public class ContentWidget : Widget
         public bool IsContentLoaded { get; private set; }
 
         /// <summary>
-        /// A token that is fired whenever the assets and scripts have loaded.
+        /// Invoked whenever the assets and scripts have loaded.
         /// </summary>
         public event Action<ContentWidget> OnContentLoaded;
 
+        /// <summary>
+        /// Invoked whenever the asset has loaded, before scripts.
```
thegoldenmule: This isn't _necessarily_ before scripts right? Since it's async?

jsundra: I believe it always will be pre-scripts. Scripts only begin to load after the asset has finished constructing (or immediately if there is no asset). The asset finish will also trigger scripts to refresh, so even if they did somehow run earlier, they'd at least update and run correctly.

PR #378, opened 2020-02-05
Mixpanel Integration

thegoldenmule:
- Added simple tracking API.
- Added MixPanel tracking implementation.
- Included mixpanel debug and release tokens.


Assets/Source/Util/UserTracking/MixPanelTrackingService.cs
```diff
@@ -0,0 +1,269 @@
...
+                }
+
+                payload["$mp_metadata"] = GenerateEventMetadata();
+
+                // send
+                _bootstrapper.BootstrapCoroutine(Send(payload));
+            }
+            catch (Exception ex)
+            {
+                Log.Error(this, "Woe is me : {0}.", ex);
```
gdamron: Ha, maybe a more descriptive message :)

Assets/Source/Util/UserTracking/TrackingEventTypes.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace CreateAR.EnkluPlayer.UserTracking
+{
+    /// <summary>
+    /// Describes all the potential event types to track.
+    /// </summary>
+    public static class TrackingEventTypes
+    {
+        public const string SIGNOUT = "signout";
```
gdamron: This doesn't appear to be used. Did you mean to create an event in `SignOutApplicationState`?

jsundra:
How do you feel about adding a method to `ITrackingService` to register/add new globals? I could see it being useful for something like the currently running appId to be registered as a global so that as we add more calls we can correlate better without having to always have the appId handy when sending an event.

Then of course `Reset` would unregister/clear the non-internal globals that the tracking service was given.

Assets/Source/Util/UserTracking/MixPanelTrackingService.cs
thegoldenmule: Whoops! I'll remove this.

Assets/Source/Util/UserTracking/TrackingEventTypes.cs
thegoldenmule: Why... yes I did!

thegoldenmule:
@jsundra The de facto mixpanel SDK has this notion, it's a good idea. I will leave it as an exercise for the next user-- when it's needed. :)

Assets/Source/Util/UserTracking/MixPanelTrackingService.cs
```diff
@@ -0,0 +1,269 @@
...
+        /// <inheritdoc />
+        public void Identify(string userId)
+        {
+            if (DistinctId == userId)
+            {
+                return;
+            }
+
+            var previousId = DistinctId;
+
```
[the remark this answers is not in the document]
thegoldenmule: Blast! No, it is supposed to be stored.

PR #380, opened 2020-02-06
Bug fix - Cannot select experience on HL when offline

gdamron:
The getter for `Task.IsComplete` throws an exception if `Task.IsFaulted` is `true` -- even if `Task.IsComplete` itself is true. Because of this, always check `Task.IsFaulted` first.

Note: In this fix, I make the assumption that if a task is cancelled, neither `onError` nor `onComplete` should be called.


Assets/Plugins/CreateAR/Commons/UwpHttpService.cs
```diff
@@ -352,18 +352,18 @@ protected void ApplyJsonPayload(object payload, HttpRequestMessage request)
...
             {
-                onComplete(responseTask.Result);
+                onError(responseTask.Exception);
             }
-            else
+            else if (responseTask.IsCompleted)
             {
-                onError(responseTask.Exception);
+                onComplete(responseTask.Result);
             }
```
thegoldenmule: Isn't there a missing `else` here such that neither callback is called?

thegoldenmule: PS - I see your comment that maybe neither is called-- but is this ensured here? I am very nervous. Perhaps at least a comment explaining to future programmers that the else is not required.

gdamron: Looking at the Task docs a bit more, `IsComplete` will be true when `IsCancelled` is true. just using an else statement here would be equivalent. I guess the real question is whether cancellation should be considered success, failure, or something else. if it is its own state, is ignoring it the right thing to do? maybe with an explicit `yield break` for clarity.

thegoldenmule: How about passing an enum in the callback, and always calling a callback (even on abort)?

gdamron: 👍

PR #382, opened 2020-02-07
Rewrote MixPanelTracking without UnityWebRequest

thegoldenmule:
Annoyingly, UnityWebRequest throws SSL cert errors on some HoloLenses and not others. Rewrote using HttpClient.

Also, added synchronous journaling so that we can make sure to resend tracking events that don't make it for whatever reason.


Assets/Source/Util/UserTracking/MixPanelTrackingService.cs
```diff
@@ -174,40 +254,111 @@ private JObject GenerateEventMetadata()
...
 
-            while (!request.isDone)
+        /// <summary>
+        /// Coroutine for waiting on a <see cref="Task"/>
+        /// </summary>
+        private static IEnumerator WaitFor(
+            Task<HttpResponseMessage> responseTask,
+            AsyncToken<string> token)
+        {
+            while (!(responseTask.IsCanceled || responseTask.IsCompleted || responseTask.IsFaulted))
```
gdamron: I strongly recommend checking `IsFaulted` first, not last

Assets/Source/Util/UserTracking/MixPanelTrackingService.cs
```diff
@@ -174,40 +254,111 @@ private JObject GenerateEventMetadata()
...
             {
-                yield return new WaitForEndOfFrame();
+                yield return null;
             }
 
-            var response = request.downloadHandler.text;
-            if (response != "1")
+            if (responseTask.IsCompleted)
```
gdamron: same here. checking `IsCompleted` can throw an exception when `IsFaulted` is true

Assets/Source/Util/UserTracking/MixPanelTrackingService.cs
thegoldenmule: Oh right! Will do

PR #383, opened 2020-02-07
Rebuild hand tracking on device recentering

jsundra:
Centralize device recenter requests. Rebuild hand tracking when the device recenters.

Do I like the name `Originer`? Not really. But I don't have a better idea either. Ideally as we dive more into handling spatialization past anchoring, I'd like to shove more into this and rename it.


Assets/Source/Application/States/Play/Hmd/Originer.cs
```diff
@@ -0,0 +1,38 @@
+﻿using CreateAR.Commons.Unity.Messaging;
+using Enklu.Orchid.Logging;
+
+namespace CreateAR.EnkluPlayer
+{
+    /// <summary>
+    /// Responsible for handling the device's centering.
+    /// </summary>
+    public class Originer
```
gdamron: Let see.

- Singularity -- the ultimate center of a galaxy
- Yogi -- for keeping the player centered
- AbsoluteZero -- can't get more zero than that

all winners...

Assets/Source/Player/Scripting/Interface/Hands/ArticulatedHandController.cs
```diff
@@ -121,12 +131,36 @@ public void GetHandData(Handedness hand, ref Dictionary<TrackedHandJoint, JointT
...
+            // Signal any existing loop to start.
+            _loopId++;
+            
+            _CoordinateSystem = Marshal.GetObjectForIUnknown(WorldManager.GetNativeISpatialCoordinateSystemPtr()) as SpatialCoordinateSystem;
+
+            UnityEngine.WSA.Application.InvokeOnUIThread(() =>
+            {
+                _spatialManager = SpatialInteractionManager.GetForCurrentView();
+
+                UnityEngine.WSA.Application.InvokeOnAppThread(() =>
```
thegoldenmule: Fwiw I've been having reliability issues with InvokeOnAppThread. I think we should prefer `ISynchronizer`.

Assets/Source/Player/Scripting/Interface/Hands/ArticulatedHandController.cs
```diff
@@ -81,26 +86,28 @@ public class ArticulatedHandController : IHandController
...
         /// </summary>
-        public ArticulatedHandController(IBootstrapper bootstrapper)
+        public ArticulatedHandController(IMessageRouter messaging, IBootstrapper bootstrapper)
         {
-            _CoordinateSystem = Marshal.GetObjectForIUnknown(WorldManager.GetNativeISpatialCoordinateSystemPtr()) as SpatialCoordinateSystem;
+            _bootstrapper = bootstrapper;
+            SetupTracking();
 
-            UnityEngine.WSA.Application.InvokeOnUIThread(() =>
+            messaging.Subscribe(MessageTypes.DEVICE_REORIGIN, (_) =>
```
thegoldenmule: Ditch the unnecessary parens around `_`.

Assets/Source/Application/States/Play/Hmd/Originer.cs
thegoldenmule: Yikes

thegoldenmule: What about `ReferenceFrame[Manager|Service|Controller]`? I think with the spaces branch it will actually eventually arrive in the anchor manager.

Assets/Source/Player/Scripting/Interface/Hands/ArticulatedHandController.cs
jsundra: 👍 Do you think `InvokeOnUIThread` is subject to the same issues you were seeing with `InvokeOnAppThread`?

thegoldenmule: That I don't know, I can't say I've used it often.

Assets/Source/Application/States/Play/Hmd/RecenterManager.cs
```diff
@@ -0,0 +1,38 @@
+﻿using CreateAR.Commons.Unity.Messaging;
+using Enklu.Orchid.Logging;
+
+namespace CreateAR.EnkluPlayer
+{
+    /// <summary>
+    /// Responsible for handling the device's centering.
+    /// </summary>
+    public class RecenterManager
```
thegoldenmule: I should have mentioned this previously, but I think we should have an `IRecenterManager` and a WSA implementation and an everything else implementation (that doesn't even publish the message). This does a few things:

- Follows the general pattern we've been using.
- Pulls #if into the module.
- Will make the _plugin_ structure easier-- i.e. there is no reason for this code to exist at all on other platforms, it's just executable bloat.

jsundra: Agreed!

Assets/Source/Application/States/Play/Hmd/PassthroughRecenterManager.cs
```diff
@@ -0,0 +1,18 @@
+﻿#if !UNITY_WSA || UNITY_EDITOR
```
thegoldenmule: This seems unnecessary.

PR #384, opened 2020-02-07
Feature/particles api

jsundra:
Most of this is from renaming `UnityParticleSystem`->`ParticleSystemJsApi` and removing the light wrapping of `IParticleSystem`. Unity's ParticleSystem and its various submodules should be handled explicitly or wrapped in a non 1:1 fashion.

Adds support for `ExternalForcesJs` which is responsible for configuring the state of a particle system's ExternalForce's module, as well as managing the list of field influencers on the module.


Assets/Source/Player/Scripting/Interface/Elements/ParticleSystemJsApi.cs
```diff
@@ -0,0 +1,235 @@
...
+                if (!field)
+                {
+                    Log.Warning(this, "No ParticleSystemForceField found on Element. ({0})", widget);
+                    return;
+                }
+
+                _module.influenceFilter = ParticleSystemGameObjectFilter.List;
+                _module.AddInfluence(field);
+                
+                Action<Element> unsub = (element) =>
```
thegoldenmule: Unnecessary parens. Please check your resharper settings, it should be showing this.

Assets/Source/Player/Scripting/Interface/Elements/ParticleSystemJsApi.cs
```diff
@@ -0,0 +1,235 @@
...
+                {
+                    _module.RemoveInfluence(field);
+
+                    if (_module.influenceCount == 0)
+                    {
+                        _module.influenceFilter = ParticleSystemGameObjectFilter.LayerMask;
+                    }
+
+                    _unsubscribes.Remove(element as ContentWidget);
```
thegoldenmule: This seems potentially dangerous. Either this should only be firing when the specific widget you want is calling it (in which case, use `widget` instead of this cast), or if this can be called for elements you are not anticipating, there is a deeper problem.

Assets/Source/Player/Scripting/Interface/Elements/ParticleSystemJsApi.cs
```diff
@@ -0,0 +1,235 @@
...
+
+            /// <summary>
+            /// Clears the list of influencers on this particle system.
+            /// </summary>
+            public void clearLimits()
+            {
+                _module.RemoveAllInfluences();
+                _module.influenceFilter = ParticleSystemGameObjectFilter.LayerMask;
+                
+                foreach (var kvp in _unsubscribes)
```
thegoldenmule: Should you also call `_unsubscribes.Clear()`?

Assets/Source/Player/Scripting/Interface/Elements/ParticleSystemJsApi.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        /// <summary>
+        /// Backing particle system.
+        /// </summary>
+        private readonly ParticleSystem _particleSystem;
+
+        /// <summary>
+        /// Mapping of schema name -> emission count.
+        /// </summary>
+        private Dictionary<ElementSchemaProp<bool>, int> _emitMap;
```
thegoldenmule: I think this should be readonly and `Clear()`'d rather than created and destroyed. This would remove the need for the null checks you have (and null checks you may be missing).

Assets/Source/Player/Scripting/Interface/Elements/ParticleSystemJsApi.cs
gdamron: +1

PR #388, opened 2020-02-12
SAI not being destroyed in the web player.

gdamron:
I missed that fact that `SceneEventHandler` is a child of `MyceliumMultiplayerController`, which is not the most appropriate place to destroy SAI. A consequence of this is that SAI is not destroyed on platforms, such as WebGL, that don't connect to Mycelium. Destroy SAI somewhere more central instead.


Assets/Source/Application/States/Play/App/AppController.cs
```diff
@@ -245,6 +249,13 @@ private void Startup(PlayMode mode)
         /// </summary>
         private void Shutdown()
         {
+            // SAI exists at the app level but behaves as a part of the scene,
+            // so destroy when exiting a scene.
+            if (_elements.ById("sai") is SaiWidget sai)
```
thegoldenmule: Should we log out an error if there is another element with id "sai"?

gdamron: 👍

PR #393, opened 2020-02-14
Implemented player score feature

shitongs:
Main features of the score JS Interface include:

-Ability to define and have multiple score categories
-Ability to add/subtract score from a category
-Ability to display incremental scores  at the location of trigger action
-Ability to set position, color, and font size of incremental scores
-Ability to assign a vine to display the total score of a category


Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+            positiveSFX = Resources.Load<AudioClip>(fileDirectory);
+
+            Data = new ScoreData
+            {
+                Name = typeName,
+                Score = 0,
+                Location = element.Schema.Get<Vec3>("position").Value
+            };
+
+            AnimationTweenData = new TweenData
```
jsundra: Could this one be renamed `PositionTweenData`?

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+        {
+            //initial opcatiy set to 0
+            Data.Color = new Col4(color.r, color.g, color.b, 0);
+
+            //update fade in/out data
+            FadeInTweenData.From = new Col4(color.r, color.g, color.b, 0);
+            FadeInTweenData.To = new Col4(color.r, color.g, color.b, color.a);
+
+            FadeOutTweenData.From = new Col4(color.r, color.g, color.b, color.a);
+            FadeOutTweenData.To = new Col4(color.r, color.g, color.b, 0);
```
jsundra: For this, to cleanup `ScoreJs` a little you can have a public method `UpdateColor` with `[DenyJsAccess]` that handles updating the dependent tweens. That way, the tweens can be made private, and the internal logic here can be nested closer to where it is used.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+        /// <param name="duration"></param>
+        /// <returns></returns>
+        public ScoreJs duration(float duration)
+        {
+            Data.Duration = duration;
+
+            //also update duration for other tween values
+            AnimationTweenData.DurationSec = duration * 2f;
+            FadeInTweenData.DurationSec = duration / 3f;
+            FadeOutTweenData.DurationSec = duration / 3f;
```
jsundra: Same here as above, this can call a `Data.SetDuration` method.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+                return;
+            }
+
+            _displayVine.schema.setString("label", Data.Score.ToString());
+        }
+
+        /// <summary>
+        /// Spawns template text with the correct values and plays positive sfx in the scene
+        /// </summary>
+        private void incrementScore(float marginScore, Vec3 location)
```
jsundra: Private methods in a JsInterface should follow our standard style conventions, so written in PascalCase.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+        {
+            //Spawn at random location within specified range
+            if(Data.RandomRange > 0)
+            {
+                float randomX = UnityEngine.Random.Range(-Data.RandomRange, Data.RandomRange);
+                float randomY = UnityEngine.Random.Range(-Data.RandomRange, Data.RandomRange);
+                location = new Vec3(location.x + randomX, location.y + randomY, location.z);
```
jsundra: Is there a reason why the Z isn't randomized here as well?

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
jsundra: Also since a negative number can be passed here, a better name would be `UpdateScore`.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+            }
+
+            //add prefix and suffix to the margin score display
+            var displayMarginScore = Data.Prefix + marginScore + Data.Suffix;
+
+            //create score text element data
+            ElementData scoreData = new ElementData
```
jsundra: more nitpicky, but since `ScoreData` is a thing, can this variable be named to something else like `scoreElementData` or `displayData`?

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+
+                Elements = new[] { scoreData }
+            });
+
+            //assign correct root parent
+            Element rootElem = _appSceneManager.Root(_appSceneManager.All[0]);
+            rootElem.AddChild(textElem);
+
+            //create audio source and play positive sfx
+            TextWidget textWidget = (TextWidget)textElem;
```
jsundra: Minor, but if `textElem` isn't needed as just an `Element`, it can be cast directly to `TextWidget` as a return from the factory creation. This way, as this grows in the future and more code is added in between, it might not be obvious that the cast is safe because the ElementData specifies TEXT as the type.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+                Root = new ElementRef
+                {
+                    Id = scoreData.Id
+                },
+
+                Elements = new[] { scoreData }
+            });
+
+            //assign correct root parent
+            Element rootElem = _appSceneManager.Root(_appSceneManager.All[0]);
```
jsundra: Use `var` as much as possible when declaring types.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+            rootElem.AddChild(textElem);
+
+            //create audio source and play positive sfx
+            TextWidget textWidget = (TextWidget)textElem;
+            textWidget.GameObject.AddComponent<AudioSource>();
+            AudioSource audioSource = textWidget.GameObject.GetComponent<AudioSource>();
```
jsundra: Same here, use `var`.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+            //start coroutine to destroy spawned score
+            _bootStrapper.BootstrapCoroutine(FadeScore(textElem, Data.Duration));
+
+            //set correct location
+            AnimationTweenData.From = location;
+            AnimationTweenData.To = location + AnimInterval;
+
+            //start tween for local animation
+            Tween animation = _tweenManager.Vec3(textElem, AnimationTweenData);
+            _tweenManager.Start(animation);
```
jsundra: Since `FadeScore` does a lot of tweening, this can be moved internal to the coroutine and it can be named something more generic like `TweenScore` or `RunTweens`.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+        /// <param name="score"></param>
+        /// <param name="duration"></param>
+        /// <returns></returns>
+        private IEnumerator FadeScore(Element score, float duration)
+        {
+            //Fade In
+            Tween fadeIn = _tweenManager.Col4(score, FadeInTweenData);
+            _tweenManager.Start(fadeIn);
+
+            yield return new WaitForSeconds(duration * 2 / 3);
```
jsundra: Would it make sense for this time to be derived from the FadeInTweenData? Looking at it, they both appear to be derived from the overall duration. It's minor, but it would be more readable if it was just `FadeInTweenData.Duration * 2`

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,339 @@
...
+
+            //Fade Out
+            Tween fadeOut = _tweenManager.Col4(score, FadeOutTweenData);
+            _tweenManager.Start(fadeOut);
+
+            yield return new WaitForSeconds(2);
```
jsundra: Can this value also be derived from the overall duration or from FadeOutTweenData like above?

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
```diff
@@ -0,0 +1,114 @@
...
+using CreateAR.Commons.Unity.Logging;
+using Enklu.Orchid;
+using CreateAR.EnkluPlayer.Util;
+using CreateAR.EnkluPlayer.IUX;
+using CreateAR.Commons.Unity.Http;
+
+namespace CreateAR.EnkluPlayer.Scripting
+{
+    [JsInterface("score")]
+    public class ScoreManagerJsInterface : JsEventDispatcher
```
jsundra: Comments throughout please :)

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        private Dictionary<string, ScoreJs> scoreMap;
```
jsundra: This can be readonly too.

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
```diff
@@ -0,0 +1,114 @@
...
+        readonly IElementFactory _elementFactory;
+        readonly IAppSceneManager _appSceneManager;
+        readonly IBootstrapper _bootstrapper;
+        readonly ITweenManager _tweenManager;
```
jsundra: These should still have their accessibility modifier on them, so `private readonly ...`

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
jsundra: And following the `_camalCase` naming convention :)

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>
+        /// Award a positive or negative score to the specified score type
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="number"></param>
+        public void award(string name, float number)
+        {
+            //if doesn't have such type, return a warning
+            if (!scoreMap.ContainsKey(name))
```
jsundra: Minor optimization, but this (and the following lookup in the `else`) can be combined with `_scoreMap.TryGetValue`.

ProjectSettings/ProjectSettings.asset
```diff
@@ -18,7 +18,7 @@ PlayerSettings:
   cursorHotspot: {x: 0, y: 0}
   m_SplashScreenBackgroundColor: {r: 0, g: 0, b: 0, a: 0}
   m_ShowUnitySplashScreen: 1
-  m_ShowUnitySplashLogo: 0
+  m_ShowUnitySplashLogo: 1
```
jsundra: Looks like this misc meta file change snuck in. Please keep the splash screen hidden :)

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
```diff
@@ -0,0 +1,114 @@
...
+            _appSceneManager = appSceneManager;
+            _bootstrapper = bootStrapper;
+            _tweenManager = tweenManager;
+            scoreMap = new Dictionary<string, ScoreJs>();
+        }
+
+        /// <summary>
+        /// Creates a score object with specified name
+        /// </summary>
+        public ScoreJs type(ElementJs element, string name)
```
jsundra: I'm not sure if it still makes sense for the creation of this to include an ElementJs reference. Since the position is optional when awarding score - maybe the default if no position is defined should be something like "2 meters infront of the user's current facing position" rather than falling back to an object that could be anywhere in the scene? What are your thoughts on that UX?

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
shitongs: You're right - z should be randomized as well since the user could be looking at it from different directions. With that, I noticed I should also let score always face camera. Will make a change to that!

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
shitongs: I think it really depends on the most common use case of how we award scores. For now, I am associating a score award related to an action taken. If that is the case, then it makes sense that the score appears around the trigger object. I could add an overload so that it defaults to 2 meters in front to handle the other use cases. What are your thoughts?

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
shitongs: What are the benefits of using var?

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
shitongs: I'm a bit confused about the public method you describe here and above. With Data.SetDuration method do you mean I should include this method within the TweenData class? Or just another public method  in ScoreJS to handle tween related updates so it is more readable?

shitongs:
> Overall, lots of great work - Nice job!! Most of the comments are style related.
> 
> I think there's one interface change that would be good though. Currently `ScoreManagerJsInterface` handles creating a `ScoreJs` but it'll persist forever. Since `ScoreJs` has methods to update its score, I don't think the manager needs to have that support anymore, and it probably makes sense to move the event dispatching directly onto the `ScoreJs` object itself.
> 
> Going back to the lifecycle comment, how do you feel about the only two methods exposed to javascript on the manager being `create` and `destroy`, that way a scripter can manage cleaning up a `ScoreJs` usage in a script's `exit` function?

I think adding the `create`  and `destroy` methods are definitely beneficial. I was just wondering if there would be cases where two different elements want to add to the same score? If that is the case then keeping the ability to update score by only invoking the name in the manager can come in handy. Since the user wouldn't need to worry about tossing the ScoreJS reference around.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
jsundra: Sorry! I read this wrong initially - I thought this was external to `ScoreJs`. Doing these calculations and whatnot in this method is fine. But seeing how it's now all in `ScoreJs` - those TweenData fields can probably be made private and readonly since the score manager doesn't need to access the data.

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
jsundra: Hmm.... Can you describe more what you mean by the "action taken"? Maybe some super short psuedo-code could help too :) I was thinking that each ScoreJs would be unique for different types of currency potentially - Money, Experience, etc which a user may be able to obtain from multiple places/objects/elements throughout the experience.

jsundra:
Hmm... I think there's definitely value in having two separate Elements modifying the same ScoreJs reference - In that case managing creation/deletion may involve pooling and reference counting to actually create/delete as needed. This kind of relates to the other thread in one of the file changes related to if element references are fully needed or not.

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
shitongs: I see! What I meant by action taken is for example if you touch a big mushroom - it gives you mushroom points every time. The points spawned would appear around the mushroom. This works well when there is one-to-one relationship between Element and ScoreJs. Or different elements trigger this score at different locations but all of those scores appear at one specific location. 
This can also be beneficial when the Element moves around - then it always has access to the latest location without requiring the user to update it.

Now I do see how it can become not as handy when different Elements affect the same score. 
I think having it default 2 meters in front is one way to solve one case. (when there is no need to associate an Element to the point awarded) 

But another remaining case is when we want scores to be spawned at different places but still contribute to the same ScoreJs. Combining what we discussed in the below thread, I think it should be handled by the score manager since it has reference to all ScoreJs. We could add a method like score.award('coins', Vec3(0,1,2), 3) to specify the specific location it needs to be spawn. 

So in pseudo code hopefully it makes more sense: 

````
// 1-to-1 relationship or many-to-1 relationship
//----------element 1---------
const score = require('score');

//when there is only one mushroom and spawns score around it
var mushroomScore = score.type(this, 'mushroom')
                             .initScore(...).color(...);


mushroomScore.award(2);     //score appears around element 1 location

//----------element 2---------
const score = require('score');
score.award('mushroom' , 2);  //score appears around element 1 location

````

````
// No elements attached
//----------element 1---------
const score = require('score');

//declare ScoreJs with no element attached
var butterflyScore= score.type('butterfly')
                   .initScore(...).color(...);


butterflyScore.award(2);     //score appears 2 meters in front of camera

//----------element 2---------
const score = require('score');
score.award('butterfly' , 2);  //score appears 2 meters in front of camera

````

````
// Each element spawn score at their corresponding place, but all contribute to same score
//----------element 1---------
const score = require('score');

//declare ScoreJs with no element attached, but specifies location
var planetScore= score.type('planet')
                      .location(this.transform.position);


planetScore.award(2);     //score appears around element 1, award points to planet score

//----------element 2---------
const score = require('score');
score.award('planet' ,this.transform.position, 2);  //score appears around element 2 location, 
                                                     award points to planet score
````


P.S. It just occurred to me we should let the user be able to specify is he/she wants to show the spawning score

**Comment by shitongs** [comment]
> Hmm... I think there's definitely value in having two separate Elements modifying the same ScoreJs reference - In that case managing creation/deletion may involve pooling and reference counting to actually create/delete as needed. This kind of relates to the other thread in one of the file changes related to if element references are fully needed or not.

With that do you mean handling deleting individually? Something like `score.deleteScore('coins');` ?
`score.clearScore('coins');` could be useful too.

File: Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs (hunk 18 above)
Comment (jsundra) [reply Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs]: Ah okay! Thanks for all the examples of the different permutations - I see what you're saying now :)

The only comment I'd have now then is that we should try and keep parity between the different calling signatures. So having something like...
```
score.award(2, [position]);
```
```
scoreManager.award('planet', 2, [position]);
```
That way the different methods of interacting with the system are closer to each other, optionals are at the end, etc.


Agreed about silently awarding too! In that case, since there definitely wouldn't be a need for a specific position, maybe a separate method with its own vocabulary would make sense. Something like `awardHidden()` or something?

Comment (shitongs) [reply Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs]: That sounds good! Will make those additions :)

**Comment by jsundra** [comment]
My comment was for the case where an experience can create 2-3 ScoreJs instances, then load into other experiences and not have those instances and their corresponding values cleared. It's a minor memory leak, but one still. But now that I think about it more, it could also lead to very hard to trace bugs where two experiences may use the same `currency` score or something and have weird side effects.

The cleanest solution, that wouldn't rely on a user's scripting would be to have a reference to `IMessageRouter` that listens to `MessageTypes.PLAY`, and clears `_scoreMap` whenever experiences are loaded. We can chat about it more offline too if you have questions about it :)

File: Assets/Source/Player/IUX/Element/ElementTypes.cs [hunk 19]
```diff
@@ -51,6 +51,9 @@ public class ElementTypes
         [ElementLink(typeof(TextWidget))]
         public const int CAPTION = 30;
 
+        [ElementLink(typeof(TextWidget))]
+        public const int TEXT = 30;
```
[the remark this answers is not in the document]
jsundra: I had actually suggested this one - Mostly because we had started removing the public facing mentions of Caption awhile ago. The docs refer to it as "Text (Caption)", vines are bound so you can use `<Caption>` or `<Text>` interchangeably. While not entirely common, we do have `ElementTypes.CONTENT` and `ElementTypes.ASSET` as aliases for each other too.

I definitely can see wanting to avoid creating more aliases though. I think renaming Caption->Text is outside the scope of this PR, so if you still advocate going that route, this PR should use Caption and we can refactor later :)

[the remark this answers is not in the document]
gdamron: Let's refactor later. I do think `TEXT` is way clearer, but I'm not a fan of the type aliases here. I'd rather have more general types.

Assets/Resources/Audio/Success_Chime.wav.meta
```diff
@@ -0,0 +1,22 @@
+fileFormatVersion: 2
+guid: 9bace33c84e475843ae8a08fafd2d315
+AudioImporter:
+  externalObjects: {}
+  serializedVersion: 6
+  defaultSettings:
+    loadType: 0
```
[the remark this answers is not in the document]
shitongs: I'm not too sure what the audio file approval process works. I connected with Haniel and he handed me those three sounds that he thinks would suit the feedback sound. Is there any documentation I can refer to for the approval process? I can definitely rename the file, change parameters, and delete the ones that aren't used!

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,451 @@
...
+            Data.Offset = offset;
+            return this;
+        }
+
+        /// <summary>
+        /// Indicates the text element that displays total score
+        /// </summary>
+        /// <param name="displayVine">ElementJs reference to display the corresponding score</param>
+        /// <returns></returns>
+        public ScoreJs display(ElementJs displayVine)
```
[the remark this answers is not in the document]
shitongs: I agree the naming can be confusing. Would `displayVine` be a better name? 

Do you mind explaining what you mean by update the vine's values?
For now, once the display vine is specified, the score on the vine will be updated and displayed appropriately.

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
```diff
@@ -0,0 +1,255 @@
...
+                //if doesn't have such type, return a warning
+                Log.Warning(this, "Warning(score.awardHidden): the specified score type does not exist");
+            }
+        }
+
+        /// <summary>
+        /// Get score of the specified score type
+        /// </summary>
+        /// <param name="name">Score type name</param>
+        public float getScore(string name)
```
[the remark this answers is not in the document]
shitongs: I agree that is a bit redundant. Will remove it :)

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
```diff
@@ -0,0 +1,255 @@
...
+            {
+                _scoreMap[name].clearScore();
+            }
+            return;
+        }
+
+        /// <summary>
+        /// Clears all elements in the map
+        /// </summary>
+        public void clearMap(object arg1, Action arg2)
```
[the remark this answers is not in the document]
shitongs: The arguments are needed to subscribe to the `IMessageRouter` so that the map clears when the session is loaded. I can add a `clearAll()` method for the users.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,451 @@
...
+            UpdateScore(marginNum, Data.Offset.Value);
+            return this;
+        }
+
+        /// <summary>
+        /// Add or substract score and let it appear at specified location
+        /// </summary>
+        /// <param name="marginNum">Marginal score awarded</param>
+        /// <param name="location">Location to spawn score</param>
+        public ScoreJs award(float marginNum, Vec3 location)
```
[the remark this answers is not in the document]
shitongs: I'm a bit confused with the benefits of adding a spawn method. Since each visualization is attached to a score increase/decrease, it feels a bit redundant to have an additional step for users.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,451 @@
...
+            OnAward?.Invoke(Data.Name, marginNum);
+            UpdateScore(marginNum, location);
+            return this;
+        }
+
+        /// <summary>
+        /// Add or substract score without spawning visualization
+        /// </summary>
+        /// <param name="marginNum">Marginal score awarded</param>
+        public ScoreJs awardHidden(float marginNum)
```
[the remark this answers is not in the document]
shitongs: Just to clarify - do you mean that it implies the user will award score without adding to the total score?

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
```diff
@@ -0,0 +1,255 @@
+﻿using System;
```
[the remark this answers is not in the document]
shitongs: Could you elaborate on separating configuration for display and animation from the data? Thank you!

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,451 @@
...
+            Data.RandomRange = range;
+            return this;
+        }
+
+        /// <summary>
+        /// Sets prefix
+        /// </summary>
+        /// <param name="prefix">Prefix String</param>
+        /// <returns></returns>
+        public ScoreJs prefix(string prefix)
```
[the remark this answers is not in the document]
shitongs: I agree that it will better for more complicated formatting! I'm curious to learn in what ways this impacts localization.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
```diff
@@ -0,0 +1,451 @@
...
+{
+    /// <summary>
+    /// Js API for score
+    /// </summary>
+    public class ScoreJs
+    {
+        /// <summary>
+        /// Score data
+        /// </summary>
+        public ScoreData Data { get; private set; }
```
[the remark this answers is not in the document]
shitongs: I referenced this structure from the `TweenData` object which is also public. I was wondering how to determine if we should to expose certain data to JS?

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
```diff
@@ -0,0 +1,255 @@
...
+            _messageRouter.Subscribe(MessageTypes.PLAY, clearMap);
+        }
+
+        /// <summary>
+        /// Creates a score object with specified name and element or returns the object with the specified name
+        /// </summary>
+        /// <param name="element">Attached Element reference</param>
+        /// <param name="name">Score type name</param>
+        /// <returns></returns>
+        public ScoreJs type(ElementJs element, string name)
```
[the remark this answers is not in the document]
shitongs: That's totally fair, will change the naming!

The concept of attaching an element might be confusing here. It is not attaching to a scoreboard location but rather to the Element's location that the visual spawned scores will be shown. For example, if the a player touches a mushroom, scores should be spawned close to the mushroom's location. If the element moves around, its score spawning location will move accordingly if it was initialized this way. I believe this method is designed originally so it stays attached to only one Element. I don't see how a score can be attached to multiple elements, since if the player calls `award`, there is no way to tell where to spawn the score. I definitely agree that multiple trigger points can contribute to the same score. For that, I think it is more intuitive to let the users call the `award(type, location)` method. 

There is definitely the counter-argument that we could just do `award(type,location)`. But there is also a use case for example that we only want the visualizations to be spawned around one specific element. Calling `award` to the same score from another element saves the user from the trouble of keeping that Element's reference and acquiring its latest position.

Assets/Resources/Audio/Success_Chime.wav.meta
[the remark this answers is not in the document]
gdamron: Usually, things get posted to #design-approvals so that non-engineers can review it and ultimately sign off

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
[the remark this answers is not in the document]
gdamron: sounds good. presumably that means this method would need a `[DenyJSAccess]` attribute as well.

Assets/Source/Player/Scripting/Interface/ScoreManagerJsInterface.cs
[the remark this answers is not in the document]
gdamron: I don't want to be too prescriptive about it, but I can imagine a few different data structures. At the top level, you could have one, two, or even three maps. Such as:

```csharp
Dictionary<string, float> _scores;
Dictionary<string, PlayerScoreDescription> _descriptions;
Dictionary<string, PlayerScoreDisplayContext> _animations;
```
or
```csharp
Dictionary<string, PlayerScoreDescription> _descriptions;
Dictionary<string, PlayerScoreDisplayContext> _animations;
```
or
```csharp
Dictionary<string, PlayerScoreEntry> _playerScores;
```
In that last case, the `PlayerScoreEntry` could be a simple class like
```csharp
private class PlayerScoreEntry
{
    public float Value;
    public PlayerScoreDescription;
    public PlayerScoreDisplayContext AnimationContext;
}
```
which would mean a `PlayerScoreDescription` class would look something like
```csharp
private class PlayScoreDescription
{
    public string Key; // key used for lookup, could default to a name if a key is not provided
    public string Name; // localizable display name
    public string Description; // localizable description
    public string DisplayString; // a localizable formatted string, such as "Wow, {0} points!"
}
```
and the animation context
```csharp
private PlayerScoreDisplayContext
{
   // color, duration, location, whatever controls are available for animation and display
}
```
You get into things like defining composite scores, where you have a class for defining how a score is actually calculated -- a total score that weights and sums other scores, for example -- but that seems like overkill right now.

Again, I'm not saying you should do exactly what I just wrote, but separating concepts in this way can make it easier for future developers to manage more complex needs in the future.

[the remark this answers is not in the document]
shitongs: Thank you for providing such detailed examples! This is definitely helped me clarify the concepts around localization :)

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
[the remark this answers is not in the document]
gdamron: There's no hard and fast rule. my thinking here is that someone could access `scoreJsObject.Data.Name` and modify it with unexpected results. While the `ScoreData` object itself is readonly, its properties are not.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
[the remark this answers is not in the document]
gdamron: maybe something like `setDisplayVine` or `setVine`? What I meant before is that, with a name like `display`, I'd expect something to display immediately when the method is called. In that case, you could take the vine and also update the label with the score.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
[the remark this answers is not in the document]
gdamron: I meant the opposite really. To me it means setting a score internally without updating the UI. In this case, the spawning +4 visualization wouldn't display, but the existing vine would, right? My suggestion is really that a more verbose name like `awardWithoutVisualization` would be easier to understand. I tend to prefer clarity over brevity.

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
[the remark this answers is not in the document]
shitongs: Sounds good!

Assets/Source/Player/Scripting/Interface/ScoreJs.cs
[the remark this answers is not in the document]
shitongs: Ok! I have changed ScoreData to private and added getters for all properties

PR #394, opened 2020-02-19
Hotfix - Object grid does not display in edit mode

gdamron:
- Open loading screen as an overlay
- Set up designer after connecting to trellis since it now checks for a connection
- Style update and null checks in UIManager


gdamron:
Thanks for the catch. With the addition of a call to `IUIElement.Revealed()` in `UIManager.Pop()`, two unit tests were failing. I've updated the tests to reflect the new behavior.

PR #397, opened 2020-02-22
JS HTTP API (Preview)

thegoldenmule:
A JS API for HTTP requests that mimics fetch.


Assets/Source/Player/Scripting/Interface/Http/AsyncHttpResponseJs.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        public AsyncHttpResponseJs(IAsyncToken<HttpResponse<string>> token)
+        {
+            _token = token;
+        }
+
+        public IAsyncResponseJs then(IJsCallback callback)
```
gdamron: comments please!

Assets/Source/Player/Scripting/Interface/Http/IAsyncResponseJs.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Enklu.Orchid;
+
+namespace CreateAR.EnkluPlayer.Scripting.Http
+{
+    /// <summary>
+    /// Interface for http responses.
+    /// </summary>
+    public interface IAsyncResponseJs
+    {
+        IAsyncResponseJs then(IJsCallback callback);
```
gdamron: comments

Assets/Source/Player/Scripting/Interface/Http/HttpJsInterface.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <param name="bootstrapper">For bootstrapping purposes.</param>
+        public HttpJsInterface(IBootstrapper bootstrapper)
+        {
+            _http = new HttpService(
+                new UTF8Serializer(),
+                bootstrapper,
+                new UrlFormatterCollection());
+        }
+
+        public IAsyncResponseJs send(string url, Dictionary<string, object> parameters)
```
gdamron: comments

gdamron: since we're not allowing headers in the parameters, we should be very clear in the documentation about what a request will/can look like

thegoldenmule: This is modeled on [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Request) but not implemented in its entirety. The reason we use a dictionary here is precisely so that we can come back and add support for a `headers` object (and later on `credentials` or `integrity` or any of the other pieces). I will make this clear in the docs.

PR #403, opened 2020-03-03
Add support for modifying a material's RenderQueue through scripting

jsundra:
Allows for the underlying RenderQueue to be changed via scripting.


gdamron:
guessing you meant to add us all as reviewers :)

jsundra:
I'm the worst :(

PR #404, opened 2020-03-03
SAI Updates

gdamron:
- Favor events over callbacks to track completion of speech, text, and focus
- Trigger focus completion when SAI is already at the target when `focus` is called.
- Expose `transform` via property
- Allow writing a sequence of texts with timeouts
- Ensure JS schema is instantiated.


Assets/Source/Player/IUX/Renderer/SAI/SaiNavigation.cs
```diff
@@ -31,7 +53,18 @@ private float DistanceToTargetSquared
                 }
             }
 
+            /// <summary>
+            /// True if the object within tolerance of a target.
```
thegoldenmule: Grammar?

Assets/Source/Player/IUX/Renderer/SAI/SaiNavigation.cs
```diff
@@ -160,17 +208,29 @@ public void Update()
             _stateMachine.Update(Time.deltaTime);
         }
 
+        public void OnDestroy()
```
thegoldenmule: Comment

Assets/Source/Player/Scripting/Interface/App/SaiJsApi.cs
```diff
@@ -262,11 +336,68 @@ private void EnsureSaiCreated()
...
+        }
+
+        /// <summary>
+        /// Invokes all callbacks in a provided list.
+        /// </summary>
+        /// <param name="callbacks"></param>
+        /// <param name="func"></param>
+        private void InvokeCallbacks(List<IJsCallback> callbacks, string func)
+        {
+            for (var i = 0; i < callbacks.Count; i++)
```
thegoldenmule: I think the list passed in needs to be copied into another list as a callback in JS may call another sai function that edits the callbacks list in some way. Were a callback added in the middle of the loop, it would not be called (because of the i) but it would be removed.

Assets/Source/Player/Scripting/Interface/App/SaiJsApi.cs
```diff
@@ -262,11 +336,68 @@ private void EnsureSaiCreated()
...
+                {
+                    callback?.Invoke();
+                }
+                catch (Exception ex)
+                {
+                    Log.Warning(this, "Error encountered in sai.write() callback: {0}", ex);
+                }
             }
+
+            _writeCallbacks.Remove(id);
```
thegoldenmule: This should be called before the callback is called.

Assets/Source/Player/IUX/Renderer/SAI/SaiNavigation.cs
gdamron: 👍

Assets/Source/Player/IUX/Renderer/SAI/SaiNavigation.cs
gdamron: 👍

Assets/Source/Player/Scripting/Interface/App/SaiJsApi.cs
gdamron: 👍

Assets/Source/Player/Scripting/Interface/App/SaiJsApi.cs
gdamron: 👍 good catch

gdamron:
Agreed @jsundra, but let's save that for the next PR. This one already has more in it than it was originally meant to 😉.

PR #405, opened 2020-03-03
Feature/storage api

jsundra:
Adds a `storage-preview` jsapi that closely follows https://developer.mozilla.org/en-US/docs/Web/API/Storage.

Example usage:
```
const storage = require('storage-preview').local;

var class = storage.getItem('class');
if (!class) {
    class = 'Tank';
    storage.setItem('class', class);
}
```

Data is stored, currently per-experience but user-independent.


Assets/Source/Player/Scripting/Interface/Storage/StandardStringStorage.cs
```diff
@@ -0,0 +1,163 @@
...
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        public StandardStringStorage(string folder)
+        {
+            _storagePath = Path.Combine(FOLDER, folder);
+            
+            try
+            {
+                Directory.CreateDirectory(_storagePath);
```
gdamron: Will creating a directory using a relative path succeed on iOS and Android? It's probably safer to use `UnityEngine.Application.persistentDataPath` + `FOLDER` + `folder`.

Assets/Source/Player/Scripting/Interface/Storage/StorageJsInterface.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// </summary>
+        public StorageJs local
+        {
+            get
+            {
+                var currentScene = _scenes.All[0];
+                if (_cachedStorage == null || _cachedStorage.Experience != currentScene)
+                {
+                    IStringStorage storage;
+#if !UNITY_WEBGL
```
thegoldenmule: This should be in EnkluPlayerModule and passed in as a dependency.

Assets/Source/Player/Scripting/Interface/Storage/WebStringStorage.cs
```diff
@@ -0,0 +1,63 @@
+﻿#if UNITY_WEBGL
```
thegoldenmule: Doesn't seem necessary.

Assets/Source/Player/Scripting/Interface/Storage/WebStringStorage.cs
```diff
@@ -0,0 +1,63 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace CreateAR.EnkluPlayer.Scripting
+{
+    /// <summary>
+    /// Web based implementation of <see cref="IStringStorage"/>. Currently session based only.
+    /// TODO: Use a jslib to actually use the browser's storage?
```
gdamron: for reference:
https://docs.unity3d.com/Manual/webgl-interactingwithbrowserscripting.html?_ga=2.136833651.2106983287.1583276798-192756895.1572458415

Assets/Source/Player/Scripting/Interface/Storage/StorageJsInterface.cs
jsundra: Heh, I was tying to avoid making a factory for such a little thing :) Will do!

gdamron: You already have the interface, so please bind by platform in `EnkluPlayerModule`

gdamron: oh, @thegoldenmule already pointed this out. will resolve mine.

Assets/Source/Player/Scripting/Interface/Storage/StandardStringStorage.cs
jsundra: Good point! I'll update it :)

PR #407, opened 2020-03-10
Reorganize the player as a Unity Package

gdamron:
This is very difficult, if not impossible to review by reading a PR. I'd suggest consuming this branch in another Unity project to get a sense of the new workflow. For the most part, the is branch just reorganizes the repository. There are a few key changes, however:

- gradle scripts are removed and will live in platform-specific repositories. see the [WebGL client](https://github.com/enklu/enkluplayer-webgl/tree/feature/to-plugin) for reference.
- Scenes have been removed and will also live in the platform-specific repositories
- Documentation has been updated for the new structure
- A new class `Runtime/Source/Application/Injection/BindableMonoBehavior.cs` has been created to allow external projects to register classes for dependency injection. This is used by RLD in the WebGL client, for example. `EnkluPlayerModule.cs`, `InjectionExtensions.cs`, and `Main.cs` are also affected by this.

For any other work in progress, also be aware that merging will become more complicated. For the most part, Git will handle changes to moved files. New files, however, will still be added to the `Assets/Source/What/Ever`. The `Assets` folder no longer exists in the new structure. Most likely, the solution will simply be to move it to `Runtime/Source/What/Ever`.


gdamron:
I should have mentioned this in the description, but I highly recommend cloning a new copy of the `enkluplayer` repo to avoid unnecessary files being import and a glut of new .meta files. There is also a getting started guide in `Documentation~/index.md`.

gdamron:
Flagging that [this commit](https://github.com/enklu/enkluplayer/pull/407/commits/234e1575a9c2260342a93a8d96e335086704d86e) removes the need for a second "PlayMode" scene. Now you just need to add the newly created `Enklu` prefab to a scene

PR #408, opened 2020-03-12
Feature/asset load progress

jsundra:
Will update the base target after `feature/to-package` merges.

Adds a white bounding box outline while an asset is loading into the player while in edit mode.

- Removed ModelOutline's usage as a selected indicator in HoloLens edit mode. It didn't work perfectly, and didn't look the best anyways.
- Changed ModelOutline to have less conflicting state, no animations, etc.
- AssetAssembler removes ModelOutline after a successful load with no errors now.
- AssetAssembler actually retries a failed asset's load now.


Runtime/Source/Player/IUX/Content/ModelOutline.cs
```diff
@@ -58,73 +52,45 @@ public ButtonWidget BtnRefresh
...
+                new Vector3(_bounds.min.x, _bounds.max.y, _bounds.max.z),
+            };
         }
 
         /// <summary>
-        /// Apply an animation effect to the bounds.
+        /// Sets the Outline to its error state.
         /// </summary>
-        public void AnimateBounds()
+        public void SetError(string error)
```
gdamron: doesn't look like you're using the `error` argument, and there's no xml comment for it

gdamron:
crap, i meant to press approved

PR #409, opened 2020-03-13
Feature/lighting fixes

jsundra:
A slew of lighting fixes!

- Ability to set pixel light count via schema (default still 0).
- Gizmo icons don't render behind the camera.
- Improved directional light gizmo.
- New spot light gizmo.
- New point light gizmo.

Queue the gifs! (Yellow: existing, Red: new. re-coloring for comparison only)

Directional light changes
Directional01
Directional02

Spot Light
SpotLight

Point Light
PointLight


Runtime/Source/Player/IUX/Gizmo/Renderers/LightWidgetGizmoRenderer.cs
```diff
@@ -88,175 +111,114 @@ private void OnRenderObject()
...
-                    ctx.Line(extents[8], extents[9]);
-                    ctx.Line(extents[10], extents[11]);
-                });
-            }*/
+            var range = _pointRangeProp.Value;
+            
+            var origin = transform.position;
+            
+            // Billboarded circle
+            Circle(origin, _cam.transform.up, _cam.transform.right, range);
```
gdamron: Definitely not a blocker, but I wonder if it's worth drawing two orthogonal circle to make it more spherical. I thing a second one just be something like  `Circle(origin, _cam.transform.up, _cam.transform.forward, range)`? Although, I guess that would just look like a line down the midle. Something to think about anyway.

jsundra: Hah! I actually had it this way at first, with circles for the transforms forward/right and was really wanting to keep it.... it's in the uncanny valley though imo. It looks great at certain angles, but others are tough. Comparing it to Unity, which darkens the parts of the sphere that would be "behind". I can share some imagery though :)

PR #412, opened 2020-03-18
Logos and loading

gdamron:
### Logos
- Remove UWP logos from this project (and move to HL project)
- Update Logo assets

### Loading
- Adjust initial offsets for camera, loading UI, and floats
- Allow UI Overlays to be "singletons", meaning only one instance of the type is allowed
- Make the loading UI a singleton in most cases, and allow it to persist across states
- Allow each design controller to define its own IUX root (resolves a `TODO`), making it possible to enable the loading screen in the web player.

### UWP build
- Fixed a few small things preventing building for HoloLens


Runtime/Source/UI/UIManager.cs
```diff
@@ -133,11 +138,34 @@ public UIManager(IUIElementFactory factory)
         /// <inheritdoc />
         public IAsyncToken<T> OpenOverlay<T>(UIReference reference, out int id) where T : IUIElement
         {
-            id = _Ids++;
+            return OpenOverlay<T>(reference, false, out id);
+        }
+
+        /// <inheritdoc />
+        public IAsyncToken<T> OpenOverlay<T>(UIReference reference, bool singleton, out int id) where T : IUIElement
```
jsundra: How would you feel about `UIReference` having an addition for singleton rather than the calling implementation? My line of thinking is that singleton nature is probably per-UI. Off the top of my head nothing jumps out as sometimes-singleton, sometimes-instanced.

With that, there'd be less chance of a bug where 1/4 `OpenOverlay` calls has the incorrect flag.

gdamron: Totally. I started out implementing it that way and changed it for "reasons". But since you've mentioned it as well, I will do it! I may end up re-naming things a bit since there's something about weird about a singleton reference.

gdamron: done

thegoldenmule: Two cents: Another potential way to do it is by creating explicit methods: `OpenOverlay` and `GetOrOpenOverlay`. That removes a bool and makes it clearer to readers.

gdamron: I almost went that way. In truth, I don't love `ThisOrThat` methods. While not perfect, I think the enum/flag approach splits the difference between readability and flexibility.

PR #415, opened 2020-03-26
Feature/sai buttons

shitongs:
Sai buttons

- the user is able to call prompt, add button options, and a callback
 - the callback will be called when a button is selected and it is able
    to indicate which one is selected
 - the user is able to adjust button text's font size, width, height, and
    alignments. The button text and sai's main text share the same font,
    shadow color, shadow offset, and line spacing


Runtime/Source/Player/IUX/Widget/SaiWidget.cs
```diff
@@ -250,6 +300,50 @@ public int Write(List<TextTransaction> script)
...
+                        _buttons[i].Label = script.ButtonOptionsText[i];
+                        //make sure button visible
+                        _buttons[i].Schema.Set<bool>("visible", true);
+                    }
+                    else
+                    {
+                        //turn rest of buttons invisible
+                        _buttons[i].Schema.Set<bool>("visible", false);
+                    }
+                }
```
jsundra: The logic flow of this if/for could be tweaked slightly to be a bit more readable:

Some abstracted example logic:
```
if (A < B) {
  for ( i, i < B ) {
    if (i < A) {
      // Do X
    } else {
      // Do Y
    }
  }
}
```
Since the for loop only runs when A < B, you can iterate over elements up til A with one action then use the other for the remainder:
```
if ( A < B) {
  for (i = 0, i < A) {
    // Do X
  }
  for (i = A, i < B) {
    // Do Y
  }
}
```

This way, the result is the same but the flow has more of a natural progression for doing the first thing, then the second, as opposed to checking which should be run for every element (even though it'd work the same).

Runtime/Source/Player/IUX/Widget/SaiWidget.cs
```diff
@@ -99,10 +141,14 @@ public class TextTransaction
         private ElementSchemaProp<float> _attractDelayProp;
         private ElementSchemaProp<float> _attractPauseProp;
         private ElementSchemaProp<float> _springProp;
+        private ElementSchemaProp<float> _buttonFontSizeProp;
```
jsundra: Other usages of fontSize use an int instead of a float - Later in the class its being cast to an int, but I don't think it's needed as a float anywhere?

Runtime/Source/Player/IUX/Widget/SaiWidget.cs
```diff
@@ -250,6 +300,50 @@ public int Write(List<TextTransaction> script)
...
+                        //initiate button
+                        _buttons.Add(CreateButton(script.ButtonOptionsText[i]));
+                    }
+                    else
+                    {
+                        _buttons[i].Label = script.ButtonOptionsText[i];
+                        //make sure button visible
+                        _buttons[i].Schema.Set<bool>("visible", true);
+                    }
+                }
```
jsundra: Similar feedback to above, this flow could also be tweaked to be more of a natural progression.

Runtime/Source/Player/IUX/Widget/SaiWidget.cs
```diff
@@ -62,6 +94,11 @@ public class TextTransaction
         /// </summary>
         private TextPrimitive _textPrimitive;
 
+        /// <summary>
+        /// SAI's buttons display
+        /// </summary>
+        private List<ButtonWidget> _buttons;
```
jsundra: Since this is being created in `LoadInternalAfterChildren` it's _probably_ safe from race conditions now. But to future proof it and make some of the code more robust, it'd be better to make this `readonly` and create it while you declare it. That way any usage doesn't have to check if it exists before checking list size or adding/removing elements.

Runtime/Source/Player/IUX/Widget/SaiWidget.cs
```diff
@@ -487,6 +625,76 @@ private void StartNextTransition()
...
+            {
+                //set values from schema
+                btn.Text.FontSize = (int)_buttonFontSizeProp.Value;
+                btn.Text.Height = _buttonHeigthProp.Value;
+                btn.Text.Width = _buttonWidthProp.Value;
+                btn.Text.Schema.Set("alignment", _buttonAlignmentProp.Value);
+            }
+
+            //set shared schemas with sai's main text
+            btn.Schema.Set("face", "camera");
```
jsundra: Minor, but `_buttonContainer` could have the face script so each child doesn't have its own.

Runtime/Source/Player/IUX/Widget/SaiWidget.cs
```diff
@@ -487,6 +625,76 @@ private void StartNextTransition()
...
+            btn.Activator.OnActivated += Button_OnActivated;
+            btn.Label = label;
+            //assign id
+            btn.Activator.Schema.Set<string>("id", (_buttons.Count + 1).ToString());
+
+            //if first button created, declare configurations
+            if (_buttons.Count == 0)
+            {
+                //buton font size
+                _buttonFontSizeProp = Schema.Get<float>("button.fontSize");
```
jsundra: Since these schema props are coming from higher up in SAI and not from the button directly, these schema props and their callbacks could be registered earlier - freeing up some of the logic here to not being "first button" specific.

Runtime/Source/Player/IUX/Widget/SaiWidget.cs
```diff
@@ -270,6 +364,14 @@ public void Animate(string name)
             _renderer.Animate(name);
         }
 
+        /// <summary>
+        /// Update button container's position to be the same as sai's
```
jsundra: Can you add a TODO in the summary to have the container be a child of SAI, and write a note or two of your findings for why this couldn't happen with the current quirks of the codebase?

shitongs:
> I'll mostly piggyback on James' comments. My one request is to move as much of the button-related stuff as possible that currently in `SaiWidget` into its own class/file. Maybe something like `PromptWidget`? It might turn out to be useful outside of SAI as well. The `SaiWidget` class is starting to get a little bloated.

To clarify, for a class like `PromptWidget`, do you mean having a class that handles the creation and activator detection of button options? Or do we want to include the question too? For now I am calling SaiWidget.Write for the question. My part only generate the necessary buttons and callbacks.

gdamron:
Up to you :), but since the new stuff is mostly concerned with buttons and callbacks, it seems reasonable for any new class to only handle those things as well.

Runtime/Source/Player/IUX/Widget/PromptWidget.cs
```diff
@@ -0,0 +1,412 @@
...
+
+        /// <summary>
+        /// Button rect change handler.
+        /// </summary>
+        /// <param name="textPrimitive"></param>
+        private void Button_OnButtonRectUpdate(TextPrimitive textPrimitive)
+        {
+            for (int i = 0; i < _buttons.Count; i++)
+            {
+                _buttons[i].GameObject.transform.localPosition = new Vector3(0f, _buttonHeigthProp.Value / 60 * BUTTON_Y_MARGIN * (i + 1), 0f);
```
gdamron, line 398: Where does `60` come from? Font size maybe?  If it's based on some dynamic property, better to use it. Otherwise, please add a comment or make it a constant.

PR #418, opened 2020-04-10
Add mesh to obj menu command

gdamron:
- Add method to convert binary mesh to OBJ string
- Add menu command to convert mesh file to obj file
- Move all editor commands from `Tools` menu to `Enklu` menu
- Rename a couple of commands that spawned warnings


Editor/Tools/Mesh/MeshCommands.cs
```diff
@@ -66,5 +66,38 @@ private static void Import()
...
+            }
+
+            var path = EditorUtility.OpenFilePanel("Open", ".", "mesh,dat");
+            if (string.IsNullOrEmpty(path))
+            {
+                return;
+            }
+
+            var bytes = File.ReadAllBytes(path);
+            var obj = new MeshExporter().ExportObj(bytes);
```
jsundra, line 92: How long does this op usually take? Might be good to throw in `EditorUtility.DisplayProgressBar` before/after without actually updating progress to at least proc a "Something is happening" feeling over a frozen editor.

gdamron: It's just a second or two for a mesh of Eden, but still no harm in adding a progress bar!

PR #422, opened 2020-05-05
Floor Height

gdamron:
Adds basic support for positioning widgets relative to the floor or ceiling. Later, this can expanded to include surfaces like tables, walls, etc. For testing, this can/should be behind Spaces, but we might as well start reviewing stuff.

- Parts of the `IMeshCaptureService` implementations have been moved to a more general `ISpatialMappingSerice` interface.
- Spatial mapping is now constant rather something that is started and stopped for capturing a mesh.
- Add Microsoft's PlaneFinding library.
- If relative positioning is enabled in an experience, the mesh in converted to planes in the background in `PlayApplicationState` and passed to `Widgets` in the scene.
- If relative positioning is enabled, Widgets's y position can be offset by the position of the floor or ceiling.


Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensPlaneCaptureService.cs
```diff
@@ -0,0 +1,305 @@
...
+                    floorPos.y = centerY;
+                    Log.Debug(this, $"Floor Height: {FloorHeight}");
+                }
+                // and the ceiling is the closest one above the camera.
+                else if (centerY > cameraY && normal.y <= -UpThreshold && area >= MinCeilingArea && yDelta < minCeilingDistance)
+                {
+                    CeilingHeight = centerY;
+                    minCeilingDistance = yDelta;
+                    Log.Debug(this, $"Ceiling Height: {CeilingHeight}");
+                }
```
jsundra, line 298: Just out of curiosity - How segmented have the planes been? I assume all the filtering is because finding "the largest horizontal one" isn't possible.

Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensPlaneCaptureService.cs
```diff
@@ -0,0 +1,305 @@
...
+                var cameraY = _camera.transform.position.y;
+                var yDelta = System.Math.Abs(cameraY - centerY);
+
+                // consider the floor the closest floor-sized plane below the camera.
+                if (centerY < cameraY && normal.y >= UpThreshold && area >= MinFloorArea && yDelta < minFloorDistance)
+                {
+                    FloorHeight = centerY;
+                    minFloorDistance = yDelta;
+                    floorPos.y = centerY;
+                    Log.Debug(this, $"Floor Height: {FloorHeight}");
```
jsundra, line 290: Should split this out and let the logger to the interpolation, especially if these are for the debug target and potentially spammy

Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensPlaneCaptureService.cs
```diff
@@ -0,0 +1,305 @@
...
+                }
+                // and the ceiling is the closest one above the camera.
+                else if (centerY > cameraY && normal.y <= -UpThreshold && area >= MinCeilingArea && yDelta < minCeilingDistance)
+                {
+                    CeilingHeight = centerY;
+                    minCeilingDistance = yDelta;
+                    Log.Debug(this, $"Ceiling Height: {CeilingHeight}");
```
jsundra, line 297: Ditto here ^^

Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensSpatialMappingService.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// </summary>
+        private void SurfaceObserver_OnDataReady(
+            SurfaceData bakedData,
+            bool outputWritten,
+            float elapsedBakeTimeSeconds)
+        {
+            OnSurfaceData?.Invoke(bakedData);
+        }
+    }
+}
```
jsundra, line 209: I like this abstraction, and I don't think it'll have merge conflicts with spaces that can't just resolve by using this implementation fully

Runtime/Source/Application/States/Play/Hmd/RelativePositionManager.cs
```diff
@@ -0,0 +1,205 @@
...
+        /// </summary>
+        public Dictionary<RelativePositionType, float> Positions
+        {
+            get
+            {
+                _positions[RelativePositionType.Floor] = FloorHeight;
+                _positions[RelativePositionType.Ceiling] = CeilingHeight;
+                return _positions;
+            }
+        }
```
jsundra, line 93: Would it be easier to use a `RelativePositions` class with each enum entry in it?

Runtime/Source/Application/States/Play/Hmd/RelativePositionManager.cs
```diff
@@ -0,0 +1,205 @@
...
+            }
+
+            var root = _scenes.Root(_sceneId);
+            if (root == null)
+            {
+                Log.Warning(this, "Cannot setup RelativePositionManager: could not find scene root.");
+                return;
+            }
+
+            _relativePosEnabledProp = root.Schema.GetOwn(PROP_ENABLED_KEY, false);
```
jsundra, line 142: Should we default this to `true`?

Runtime/Source/Application/States/Play/Hmd/RelativePositionManager.cs
```diff
@@ -0,0 +1,205 @@
...
+
+            var all = _elements.All;
+            for (var i = 0; i < all.Count; i++)
+            {
+                var element = all[i];
+                if (element is Widget widget)
+                {
+                    widget.UpdateWorldOffset(Positions);
+                }
+            }
```
jsundra, line 169: Maybe this should be a pull Widgets do instead of a push? That way there isn't so much casting every frame, and non-visible widgets won't need to update.

Runtime/Source/Application/States/Play/Hmd/RelativePositionManager.cs
```diff
@@ -0,0 +1,205 @@
...
+            _planes = planes;
+            _elements = elements;
+
+            _positions[RelativePositionType.World] = 0f;
+        }
+
+        /// <summary>
+        /// Starts plane finding if enabled for the current scene.
+        /// </summary>
+        public void Start()
```
jsundra, line 126: Since `Start` is called from `PlayApplicationState` which is grabbing the scene/root for the quality manager, these checks could be surfaced there and you could call Start off of a specific passed in root. Not sure if there's a huge benefit other than this class not being reliant on `IAppSceneManager`

Runtime/Source/Application/States/Play/Hmd/RelativePositionManager.cs
```diff
@@ -0,0 +1,205 @@
...
+                return;
+            }
+
+            var all = _elements.All;
+            for (var i = 0; i < all.Count; i++)
+            {
+                var element = all[i];
+                if (element is Widget widget)
+                {
+                    widget.UpdateWorldOffset(Positions);
```
jsundra, line 167: `_positions` is only updated as a side-effect of `Positions` being queried. I'd update `_positions` at the start of `Update` and pass that instead of hitting the get accessor of `Positions` multiple times per frame

Runtime/Source/Player/IUX/Widget/Widget.cs
```diff
@@ -336,6 +343,36 @@ public void BringToTop()
...
+                    vec.y = positions[type];
+                    break;
+                }
+            }
+
+            if (vec != WorldOffset)
+            {
+                WorldOffset = vec;
+                UpdateTransform();
+            }
```
jsundra, line 373: Similar to comments in `RelativePositionManager`, if this is switched to pulling from the manager, then the prop updating can control whether a Widget pulls relative positioning in its update, reducing the amount of noops if a Widget is set to World.

Runtime/Source/Player/IUX/Widget/Widget.cs
```diff
@@ -364,6 +401,10 @@ protected override void LoadInternalBeforeChildren()
             _alphaProp = Schema.GetOwn("alpha", 1f);
             _alphaProp.OnChanged += Alpha_OnChanged;
 
+            _relativePosProp = Schema.GetOwn(
+                "relative.position",
```
jsundra, line 405: You'll see in the upcoming spaces PR that I'm running into this same issue too of having instances and managers loosely sharing schema keys, or sometimes sharing them off of referencing a const on the manager. Both are kind of cringey. We have the element schema defaults, but I'm starting to think we need something a level above that, which is just a list of types/keys, similar to `MessageTypes`. That way as we change things and deprecate schema, we could also avoid re-using old keys and causing some experiences to have weird bugs potentially.

Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensPlaneCaptureService.cs
```diff
@@ -0,0 +1,305 @@
...
+                Log.Warning(this, "Start() called but plane capture service is already running.");
+                return;
+            }
+
+            Log.Debug(this, "Starting plane capture service.");
+
+            _isRunning = true;
+
+            _floor = GameObject.CreatePrimitive(PrimitiveType.Plane);
+            _floor.GetComponent<Renderer>().material = _config.SurfaceMaterial;
```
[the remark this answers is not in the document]
gdamron: nope, just a mistake 🙂

Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensPlaneCaptureService.cs
```diff
@@ -0,0 +1,305 @@
...
+
+                meshes.Add(new MeshData
+                {
+                    Vertices = mesh.vertices,
+                    Normals = mesh.normals,
+                    Indices = mesh.triangles,
+                    Transform = filter.transform.localToWorldMatrix
+                });
+
+                yield return null;
```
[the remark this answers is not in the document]
gdamron: That's sort of left over but definitely needs some attention. The reason it's there is because recalculating the normals and bounds can be expensive. But it should be checking if it's taken too long instead of `yield returning null` each iteration.

Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensSpatialMappingService.cs
gdamron: phew, I was worried about that

Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensPlaneCaptureService.cs
gdamron: 👍

Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensPlaneCaptureService.cs
gdamron: 👍

Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensPlaneCaptureService.cs
gdamron: I started out that way, but the problem is that the largest plane might not be the floor you want. Like, the HL might pick up the floor of the next story down or the ground outside of your house. That's why I settled on "the closest plane below you large enough be considered the floor"

Runtime/Source/Application/States/Play/Hmd/RelativePositionManager.cs
gdamron: I did this, knowing it'll need to change down the line, because we don't have a clear enough use case for other types of relative positions. I'll give it another pass. At the very least, I'll ad a comment.

Runtime/Source/Application/States/Play/Hmd/RelativePositionManager.cs
gdamron: Figured I'd start with `false` until it's proven itself in the wild. Consider it an experimental feature for now.

Runtime/Source/Application/States/Play/Hmd/RelativePositionManager.cs
gdamron: I'll give it another pass. But the initial attempt at pulling let to a big fat circular dependency.

Runtime/Source/Application/States/Play/Hmd/RelativePositionManager.cs
gdamron: Yep, good idea.

Runtime/Source/Application/States/Play/Hmd/RelativePositionManager.cs
gdamron: 👍

Runtime/Source/Player/IUX/Widget/Widget.cs
gdamron: Yeah, I've been thinking that too. The flexibility of `Schema` is one of our great strengths, but we've let it get a little unruly in some applications.

Runtime/Source/Application/States/Play/Hmd/MeshCapture/Service/HoloLensPlaneCaptureService.cs
```diff
@@ -0,0 +1,305 @@
...
+                var plane = planes[i];
+                Planes.Add(plane);
+
+                var area = plane.Area;
+                var bounds = plane.Bounds;
+                var center = bounds.Center;
+                var normal = plane.Plane.normal;
+                var centerY = center.y;
+
+                var cameraY = _camera.transform.position.y;
```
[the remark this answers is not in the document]
gdamron: 👍

gdamron:
closing since there will be a fair amount of integration work and since requirements have likely changed.

PR #423, opened 2020-05-07
Feature/spaces 2.0

jsundra:
# Spaces
source

#### What is Spaces?
Spaces is the ability for a user to build an experience, without having to deal with 90% of the steps required for anchoring in the past. When in edit mode, a user just has to use the edit menu to select Anchoring -> Link new space. The system will scan and create anchors for the user. Scene content is automatically anchored, removing the old requirement of experience designers having to nest content under an Anchor Element.

To achieve this, Spaces exist as a separate Widget that contains the scan data and anchors as children that is loaded separately from a Scene. The scene graph is left intact, but children of a scene are reparented during Play/Edit mode to be under anchors from the Space. Whether a Space or Scene loads/unloads, all reparenting is reversed, and no scene graph changes are propagated back to Trellis.

## Major Changes
- Spaces and Anchoring related code is now under a `Spaces` namespace.
- `SpaceManagementState` is a new state for creating, editing, and deleting spaces - both the scan and anchors.
- `IPrimaryAnchorManager` is completely obliterated. `IAnchorManager` is responsible for importing/exporting anchor data, while `IAnchorLocator` now handles tracking loaded anchors and managing the prior relative fallback system. Other responsibilities are distributed elsewhere.
- Anchor state has been separated. There is now a separate distinction for an individual anchor's locatability
  - `LocatingStatus`
    - Unavailable: The anchor is new, or has not downloaded/imported data yet.
    - NotLocated: Anchor available, but not able to find its position in the world.
    - Located: Anchor available, and location found.
  - `AnchoringState`
    - None: The default state.
    - Downloading/Importing: Data is being ingested from Trellis.
    - Exporting: Data is exporting to Trellis.
    - Ready: Anchor data is available for anchors to use.
    - Error: Something bad happened.
- Anchors no longer import/export individually. They load/export as part of a single batch to reduce memory usage. Existing anchors will still work under a legacy code flow.
- `ElementControllerGroup`s work off all Elements, not just a scene root.

#### Experience Edit Mode
Experience - Edit Mode
#### Space Management States
Space Management States

#### SpaceLinkManager
Systems - SpaceManager
#### Spaces
Systems - Spaces
#### Anchoring
Systems - Anchoring

### Notifications
User-facing notifications can be sent via `INotificationManager`. These are throttled and will display as small text that scrolls past the user's view.
### Mesh Occlusion
`ScanWidget` downloads the mesh scan in all modes now. An occlusion shader is used on the mesh in play/edit mode to increase an experience's immersion. This will default on, but be able to be turned off via the web editor.
### MaterialManager & PostProcessing
`IMaterialManager` has been rewritten to support overriding a Widget's material at runtime based on a known material from `MaterialTypes`. Similarly, shaders can be overwritten on materials. The existing usage of `IMaterialManager` has been converted, and the ScanWidget now uses this feature to toggle the mesh between displaying as occlusion vs its scan visualization while in space management.
`MainCamera` has an added hook to surface `OnRenderImage`, allowing some basic post processing effects to be used.

## Improvements
- Async.dll updated to support `AsTask`
- Http.dll updated to support non-UWP `PutFile`/`PostFile` support.
- `MeshCaptureExportServiceWorker` uses a ConcurrentQueue under the hood now.
- Added more error information for DI issues during development.

## Minor Changes/Additions
- Removed unused/broken items from Edit mode's menu.
- `ElementSplashController`s can display status via their UIs.
- Confirmation UI is a bit more centered.
- `ReparentUtils` created to abstract some Element reparenting maths.
- Additional ElementControllerFilters created.
- Floats don't default to showing their focus.
- Mixpanel no longer attempts to send metrics when it has no token.

## Bugfixes
- Double cursor in experience list is removed.
- Hidden buttons no longer activate when their voice command is recognized. The recognizer will still play a bloop however.
- ElementControllers are actually destroyed now.

## TODO
- Final shaders for the mesh, for use in WebGL & Space Management mode.
- Updated anchor visualizations.
- The inevitable slew of bugs.


Runtime/Source/Player/Spaces/SpaceLinkManager.cs
```diff
@@ -0,0 +1,338 @@
...
+using CreateAR.Commons.Unity.Async;
+using CreateAR.EnkluPlayer.App;
+using CreateAR.EnkluPlayer.IUX;
+using Enklu.Data;
+using Enklu.Orchid.Logging;
+using Void = CreateAR.Commons.Unity.Async.Void;
+
+namespace CreateAR.EnkluPlayer.Spaces
+{
+    public class SpaceLinkManager : ISpaceLinkManager
```
gdamron, line 14: comment

Runtime/Prefabs/Vines/Menus/Design/Design.MainMenu.txt
```diff
@@ -74,10 +71,10 @@
 				<Button id='btn-exp-duplicate' label='Duplicate' icon='camera' voiceActivator='duplicate' />
 			</SubMenu>
 			
-			<SubMenu label='Application' icon='text' voiceActivator='application' >
-				<Text id='txt-version' label='' fontSize=80 width=200.0 />
```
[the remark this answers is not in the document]
jsundra: 👍 I'll add it somewhere more obvious, yeah it even came up yesterday again so it's worthwhile to throw in :)

Runtime/Prefabs/Vines/Menus/Design/Anchoring.Unlocated.txt
```diff
@@ -0,0 +1,18 @@
+<?Vine>
+
+<Float>
```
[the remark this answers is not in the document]
jsundra: This one should be good without it, floats also rotate to face the user. Sometime it's added to a child Container if a float has multiple children and needs better pivot points for the facing angle

Runtime/Source/Player/Spaces/StandardSpaceController.cs
```diff
@@ -0,0 +1,326 @@
...
+        
+        /// <summary>
+        /// Contextual name for a new scene.
+        /// </summary>
+        /// <returns></returns>
+        private string DefaultSpaceName()
+        {
+            if (string.IsNullOrEmpty(_app.Name))
+            {
+                return $"Space_{StringExtensions.RandomIdentifier(4)}";
```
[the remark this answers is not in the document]
jsundra: Definitely doesn't need to be unique. It's more of a flow safety than a storage one. Just incase someone is using a build that doesn't have a default experience, and they enter space management before going into an app? Most likely not going to happen in practice, but figured it's better to have _something_. But that something can just be "New Space". The random identifier was a placeholder from before the app name was used

Runtime/Source/UI/Common/Hmd/LoadingUIView.cs
```diff
@@ -15,6 +17,14 @@ public class LoadingUIView : MonoBehaviourIUXController, ICommonLoadingView
         public TextWidget TxtStatus { get; set; }
         [InjectElements("..btn-abort")]
         public ButtonWidget BtnAbort { get; set; }
+        [InjectElements("..txt-version")]
+        public TextWidget TxtVersion { get; set; }
```
gdamron, line 21: can this be more hidden? while having the version is occasionally useful for _us_, it's an ugly addition to something users see very often.

jsundra: This actually came in as a request so we could help triage what version a client might actually be running. I'll dig a little for some alternatives, it'd be great if we could have the version in the window title but no where else. In the very least, I can make that text smaller if it ends up staying in the loading popup :D

jsundra: I looked into it a bit. Unfortunately setting the ApplicationView's title property doesn't seem to do anything, both in C# and in the C++ project wrapper. So for now, I just made the version much smaller so it's actually a small focus and not as big as the loading text!

PR #424, opened 2020-05-11
IUX fixes

gdamron:
A bunch of small changes related to IUX
- Updated Color palette to reflect Brand Colors (they will probably need to be tweaked more, but they're a start)
- Updated fonts to be consistent with Enklu Brand
- Updated Internal vines to use a more consistent style.
- Applied a bunch of existing changes on prefabs to their base prefabs
- Set the Camera start position to 0,0,0 on UWP to prevent the loading screen from starting behind the user.
- Make the play button close-able
- Clarified labels on some floats
- Add `icon.*.color` schema props to enable colorizing icons
- Add 'fov.distance` schema props to float to allow setting a max distance before following
- Made logging floats essentially stationary (which appears to be the original intent) by applying a large max distance.


Runtime/Source/Player/Vine/ElementDescriptionListener.cs
```diff
@@ -342,36 +342,16 @@ public void ExitAttributeValue(VineParser.AttributeValueContext context)
...
-                if (!float.TryParse(substrings[0], out r)
-                    || !float.TryParse(substrings[1], out g)
-                    || !float.TryParse(substrings[2], out b))
-                {
-                    throw new Exception(string.Format(
-                        "Col4 could not be parsed for attribute {0} : {1}.",
-                        name,
-                        GetExceptionLocation(context)));
-                }
-
```
jsundra, line 364: Hey okay you're going to love this one.... Sooooo `ElementDescriptionListener` is actually maintained very poorly from another repo... and this version isn't even the master branch of that repo.... The latest of this matching up with: https://github.com/enklu/create-commons-vine/blob/feature/parsing-improvements/Vine.Tests/ElementDescriptionListener.cs.

Maybe it's fine to finally break away from having enkluplayer's parsing come from the test folder of the vine repo though?

gdamron: Oh, it was originally test code?! That explains a lot. I am 100% fine with breaking away from that code.

PR #426, opened 2020-06-17
Fix for Gaze API

gdamron:
Lazily asks the user for permission to track eye gaze


Runtime/Source/Player/Element/Gaze/GlanceController.cs
```diff
@@ -32,7 +32,13 @@ public GlanceController()
         {
             _CoordinateSystem = Marshal.GetObjectForIUnknown(WorldManager.GetNativeISpatialCoordinateSystemPtr()) as SpatialCoordinateSystem;
         }
-        
+
+        /// <inheritdoc />
+        public void RequestAppPermission()
+        {
+            _ = EyesPose.RequestAccessAsync();
```
jsundra, line 39: `_` snuck in from testing? Or is this some fancy thing in the latest C# spec that wouldn't throw in a compilation error?

gdamron: fancy C# thing 🙂

gdamron:
Doesn't hurt performance, but that doesn't mean we shouldn't check whether it's supported or previously authorized. At the very least, it would allow us to short circuit the processing.

gdamron:
Double checked, and `RequestAccessAsync` is the only way to check whether permission has been granted. For now, I'll add a check for `IsSupported` as well

PR #427, opened 2020-06-18
Add login window to Unity

gdamron:
This revives and existing account management window for the Unity editor. I mainly cleaned it up, updated references, and made it look less weird.


Editor/Tools/Application/EditorConfigurationManager.cs
```diff
@@ -12,10 +12,10 @@ public class EditorConfigurationManager
...
         /// </summary>
         /// <returns></returns>
-        private static string APP_CONFIG_DIRECTORY = "Packages/com.enklu.player/Runtime/Resources";
-        private static string APP_CONFIG_NAME = "ApplicationConfig.json";
-        private static string USER_CONFIG_DIRECTORY = Path.Combine(UnityEngine.Application.persistentDataPath, "Login");
-        private static string USER_CONFIG_NAME = "LoginConfiguration.local";
+        internal static readonly string UserConfigDirectory = Path.Combine(UnityEngine.Application.persistentDataPath, "Login");
+        private const string AppConfigDirectory = "Packages/com.enklu.player/Runtime/Resources";
+        private const string AppConfigName = "ApplicationConfig.json";
+        private const string UserConfigName = "LoginConfiguration.local";
```
jsundra, line 18: More of the codebase uses `ALL_CAPS` for consts than not, right?

Editor/Tools/PersistentDataHooks.cs
```diff
@@ -36,6 +37,12 @@ public static void ClearAllPersistentData()
             {
                 try
                 {
+                    // but don't log out.
+                    if (EditorConfigurationManager.UserConfigDirectory == directory.FullName)
+                    {
+                        continue;
+                    }
```
jsundra, line 44: Could there be a separate button to clear credentials? I know there's times during testing where I start the app just to logout. Plus, it'd be good for parity's sake.

Editor/Tools/Application/EditorConfigurationManager.cs
gdamron: that's what Re/Sharper commands for private constants and non-private static strings. I'm certainly not wedded to it, but I tend to follow Re/Sharper since it's approaching standard.

Editor/Tools/PersistentDataHooks.cs
gdamron: there is!

Editor/Tools/Application/EditorConfigurationManager.cs
jsundra: Sorry, are you saying the resharper setting is for all caps or pascal case? Mine right now is set for all caps, but it might have just been part of going with the existing styling.

I'm not particularly set on either one either, but I think I prefer public/private to be styled the same for consts. Mostly because the weight of changing accessibility is harmless for consts compared to changing a field.

Regardless, I don't think I have a huge preference, but if we are going to start transitioning towards a new style we should be consistent across the PR. I think I spotted a private const somewhere further in the review that was a different style

Editor/Tools/EditorWindow/Toolsuite/Login/LoginView.cs
```diff
@@ -20,15 +21,10 @@ public class LoginView : IEditorView
...
-        /// <summary>
-        /// Position.
-        /// </summary>
-        private Vector2 _scrollPosition;
-
         /// <summary>
         /// Title for signin.
         /// </summary>
-        private const string SIGNIN_PROGRESS_TITLE = "Connecting to Trellis...";
+        private const string SIGNIN_PROGRESS_TITLE = "Signing in to Enklu Cloud...";
```
jsundra, line 27: Yah here's the line :)

gdamron: It's funny, changed the style because I thought I didn't have string feelings and wanted to make the squiggles go away. Now that it comes down to, I do have strong feelings -- `UPPERCASE_SNAKE_FOR_ALL_CONSTS`! Screw you Re/Sharper.

PR #428, opened 2020-06-18
Player metrics pass

gdamron:
- Adds ability to send metrics across the web bridge
- Updates metrics keys to new style
- add tracking for QR login, email login, and script module usage


Runtime/Source/Async/Async.cs
```diff
@@ -47,11 +47,22 @@ public TimedOutException() : base("Operation timed out.")
...
 
             // wait
-            bootstrapper.Wait(milliseconds, () => timedToken.Fail(new TimedOutException()));
+            bootstrapper.Wait(milliseconds, () =>
+            {
+                if (!didSucceed)
+                {
+                    timedToken.Fail(new TimedOutException());
+                }
+            });
```
jsundra, line 65: 🏆

Runtime/Source/Util/UserTracking/BridgeTrackingService.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                Metric = new TrackingBridgeEventMetric
+                {
+                    Event = type,
+                    Payload = parameters
+                }
+            }, out var bytes);
+
+            _bridge.Send(Encoding.UTF8.GetString(bytes));
+        }
```
jsundra, line 76: How do you feel about wrapping this in a try/catch? Ideally it never fails, but I'd hate for code flow to blow up because a silly failure like metrics :)

gdamron: 👍 will do

PR #429, opened 2020-07-02
Feature/offline multiplayer

jsundra:
Attempts to build an element via multiplayer. On failure (or while not connected), the API creates it locally anyways.

Not perfect, in the future will need to sync the new Elements back to the server after a connection is established, etc etc. But for now, prevents a lot of fallback code being needed in an experience's scripting layer.


Runtime/Source/Player/Scripting/Interface/Multiplayer/MultiplayerJsInterface.cs
```diff
@@ -183,9 +190,48 @@ public ElementBuilderJs builder(ElementJs parent)
...
+            else
+            {
+                mpCreate = new AsyncToken<Element>(new Exception("No active multiplayer connection"));
+            }
+
+            mpCreate
+                .OnSuccess(rtnToken.Succeed)
+                .OnFailure(exception =>
+                {
+                    // TODO: Sync elements created offline with the server on reconnect
```
gdamron, line 211: I think offline creation should be moved into the multiplayer controller since client side prediction is a basic best practice for multiplayer anyway. In the very short term, I think it's ok to keep it as is, but there are important improvements to make: We should always create locally first and reconcile against what we receive from the server a little later. This may require some changes to the messages we send -- I'm not sure if we send a transaction id for example -- but it shouldn't be too bad. It's probably worth assessing the level of effort for doing this now. Much of this applies to our tcp connections to Trellis too, btw, although I'm not suggesting that for this PR. That's what [#3960](https://enklu.tpondemand.com/entity/3960-schema-changes-get-overwritten) is ultimately about.

But also, I think our approach to multiplayer has a fundamental flaw. We should be multiplayer by default. That means I'd like to keep this API as thin as possible with a longer term goal of deprecating it, at least in it's current form.

jsundra: Definitely 👍 I'll dig into it a bit this week and see how adventurous the first pass of having the clients create things would be. I want to say it won't be too bad. Plus, it would also lead well into clients being able to delete networked objects in a much better way than possible currently.

jsundra:
Closing until I have more time to do this properly.

PR #433, opened 2020-07-16
SAI Callback fixes

jsundra:
SAI text completion events invoke with the id of transactions that finish, while the `Write` function was returning the queue length (which often was -1 since it was processing mid invoke as well).

While looking into this, I had originally thought it was a jint method overload invocation order issue. So I did some testing there and found out some of our earlier issues are no longer valid. Overloads with different parameter counts are okay, but there are still problems with optional parameters and methods with overloads of different types. That.... wasn't a quick fix so I didn't go down that path.

I was on the fence of adding legacy fallbacks for these, but the snap, edit, and capture apis I'm pretty sure wouldn't have a lot of external usage to date. Happy to add them in if we want them though.


Runtime/Source/Player/IUX/Widget/SaiWidget.cs
```diff
@@ -274,13 +274,7 @@ public int Write(List<TextTransaction> script)
...
             StartNextTransition();
 
-            if (_textTxnQueue.Count == 0)
-            {
-                return -1;
-            }
-
-            var last = _textTxnQueue[_textTxnQueue.Count - 1];
-            return last.Id;
+            return script[script.Count - 1].Id;
```
gdamron, line 277: The script can be empty, and we should guard against it. Maybe catch it in `SaiJsApi` and pass the error into the callback?

jsundra:
Double checked the docs, they already correctly mention that the callback is invoked after the last entry is run in a script with multiple entries

